PR #2, opened 2017-05-17
Pull request for mircea

FlyingBird95:


mircealungu:
eheh. title of the pull request should be about what's inside not for whom it is ;)

flask-dashboard/dashboard/routings/setup.py
```diff
@@ -0,0 +1,63 @@
...
+        set_setting('username', form.username.data)
+        if form.password.data:
+            set_setting('password', form.password.data)
+
+    return env.get_template('settings.html').\
+        render(link=link, session=session, version=version, database_name=database_name, user_var=user_var,
+               form=form, old_password=old_password)
+
+
+@user_app.route('/' + link + '/rules', methods=['GET', 'POST'])
```
mircealungu: i think the code would be cleaner if, instead of adding the new routes to the app, you would create a new blueprint, and install that blueprint in the app. 

see: http://flask.pocoo.org/docs/0.12/blueprints/

mircealungu: Folks, please either:
1. address the afore-mentioned comments
2. create issues if there is something that you can't tackle right away. 

once this is done, let me know, and i'll merge the pull request.

PR #3, opened 2017-05-29
Development

FlyingBird95:
Implemented the following changes:
- Resolved issues with installing. It should be possible now to install using python (v2.7) and python3.
- Implemented blueprint for the application. The dashboard is now added to a blueprint, which is registered on the application.
- Simplified graphs. In a previous version, the graphs were too complex. Now, it shows the same amount of data in a different visual content, which makes it easier to see changes.
- Implemented a config-object. The config-object can easily be changed by a config-file or by modifying the attributes in this object.


mircealungu:
Trouble in paradise. 

Have a look at: 

      https://zeeguu.unibe.ch/api/dashboard/

You will see that all the links are broken with the exception of the /rules, which I've rewritten using url_for. Have a look at this commit:

    https://github.com/mircealungu/automatic-monitoring-dasboard/commit/2fc8dea7ee85dbc5bcfab2e8a55e4a673e986c70

For more info please read up on : 

    http://flask.pocoo.org/docs/0.12/blueprints/#building-urls

Also, you have the same problem with the static files. They must be accessed using url_for not with hardcoded urls. See: 

    http://flask.pocoo.org/docs/0.12/blueprints/#static-files

Finally, I see a flask-dashboard.db file created. I see it modified when the endpoints are accessed, but the data on https://zeeguu.unibe.ch/api/dashboard/rules never changes... 

Can you address these issues? 
Thanks!

mircealungu:
Actually, we don't have to close the request. If you create new commits on the `development` branch, they'll be added automatically to this pull request anyway.

mircealungu:
nice! the last commit fixed the urls and the static assets. 
there is still not data displayed: https://zeeguu.unibe.ch/api/dashboard/rules

we have to investigate why.

PR #18, opened 2017-06-06
Moved setup.py to the root of the repo.

kloostert:


setup.py
```diff
@@ -23,16 +23,15 @@ def tuple_list(directory):
     README = ''
 
 # Copy all data_files from the 'static'- and 'templates'-folder into their destination
-data_files_list = tuple_list('static')
-data_files_list.extend(tuple_list('templates'))
-# TODO: also add config-file to data_files_list
+data_files_list = tuple_list('flask-dashboard/static')
+data_files_list.extend(tuple_list('flask-dashboard/templates'))
```
mircealungu: guys, here you're reinventing the wheel. 
you can use MANIFEST.in to declare the static files that must be installed in the package. 

see this: https://github.com/mircealungu/Zeeguu-Web/blob/master/MANIFEST.in

mircealungu: see also: https://docs.python.org/2/distutils/sourcedist.html#manifest-template

FlyingBird95: When creating the setup script, I also looked into the manifest.in file, but I couldn't figure out how to declare the static files. This function contains the same functionality

mircealungu: i looked a lot for the solution too, but it's too lines of manifest file, as the first link shows. it would be nice if you could try to replace the many lines with the few.

kloostert: We've created a manifest file, and changed the setup file accordingly. 
We hope it will work with Travis now.

PR #40, opened 2017-09-23
Multiple guest passwords, plot updates, improved security against guests downloading data

kloostert:


dashboard/routings/result.py
```diff
@@ -197,7 +198,8 @@ def get_hits_per_hour(end):
 def get_time_per_version_per_user(end, versions):
     user_data = {}
     data = [t.execution_time for t in get_all_measurement(end)]
-    average = sum(data) / len(data)
+    # compute the average for determining the default size
+    average = math.sqrt(sum(data) / len(data))/1250
```
mircealungu: magic constant!

dashboard/security.py
```diff
@@ -36,18 +36,15 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-def security(admin_only=False):
-    def decorator(f):
-        print(admin_only)
-        return f
-    return decorator
+def is_admin():
+    return session and session.pop(config.link + '_admin', False)
```
mircealungu: looking only at these two lines, it seems to me that you would use `peek` or `top` instead of `pop` which modifies the stack. or?

dashboard/routings/result.py
kloostert: It is now defined as a constant at the top of the file, is this what you had in mind?

dashboard/security.py
kloostert: You're completely right, this was a bug; it should be fixed now.

PR #46, opened 2018-01-23
Create CODE_OF_CONDUCT.md

FlyingBird95:


mircealungu:
This feels strange because to me all that's said there goes without saying... 
We should not have to write down in a file the fact that we'll be nice to each other. 

Is it only me? @vandriko ?

PR #47, opened 2018-01-23
Create LICENSE

FlyingBird95:


mircealungu:
I normally prefer MIT. 
Is there a good reason for GPL-ing the code? 
@vandriko what do you think?

vandriko:
Apache2? It's the one we usually use to avoid the GPL restrictions...

mircealungu:
@FlyingBird95 @vandriko  @kloostert  - do you people agree with using MIT? Or you have a strong reason for going with a viral license?

vandriko:
Fine with me...

FlyingBird95:
MIT is also fine with me

kloostert:
Yes, same here.

PR #79, opened 2018-04-05
one possible way

mircealungu:
useful for testing with / without the dashboard, like this: 

```
dashboard.config.monitor_everything = True
dashboard.bind(app=app, blue_print=api)

```


mircealungu:
@FlyingBird95  - we don't need this PR anymore; it's redundant with your `outliers_enabled` config option,right?

PR #84, opened 2018-04-08
resolved bug

FlyingBird95:


flask_monitoringdashboard/database/function_calls.py
```diff
@@ -27,8 +27,10 @@ def add_function_call(time, endpoint, ip):
     """ Add a measurement to the database. """
     with session_scope() as db_session:
         group_by = None
-        if config.get_group_by:
+        try:
```
mircealungu: i would still keep the conditional.
otherwise, there will always be an error, even for those 
cases where people don't install group_by

flask_monitoringdashboard/database/function_calls.py
```diff
@@ -27,8 +27,11 @@ def add_function_call(time, endpoint, ip):
     """ Add a measurement to the database. """
     with session_scope() as db_session:
         group_by = None
-        if config.get_group_by:
-            group_by = config.get_group_by()
+        try:
+            if config.get_group_by:
```
mircealungu: the last think i'll ask on this PR: 
please reverse the order: if config then try catch.

flask_monitoringdashboard/measurement.py
```diff
@@ -70,8 +70,8 @@ def wrapper(*args, **kwargs):
         add_function_call(time=t, endpoint=endpoint, ip=request.environ['REMOTE_ADDR'])
 
         # outlier detection
-        endpoint_count[endpoint] += 1
-        endpoint_sum[endpoint] += t
+        endpoint_count[endpoint] = endpoint_count.get(endpoint, 0) + 1
+        endpoint_sum[endpoint] = endpoint_sum.get(endpoint, 0) + t
```
mircealungu: cool.
how did it work till now?

PR #96, opened 2018-04-30
updated all database calls by combining the sessions

FlyingBird95:
Updated the code of the database by combining the sessions.

First, the code was like: 
screen shot 2018-04-30 at 16 26 17

Now, the code looks like:
screen shot 2018-04-30 at 16 26 54

As a result, the process of loading the pages will go faster

### DIFF
--- a/flask_monitoringdashboard/core/measurement.py
+++ b/flask_monitoringdashboard/core/measurement.py
@@ -7,6 +7,8 @@ import datetime
 from flask import request
 from functools import wraps
 from flask_monitoringdashboard import user_app, config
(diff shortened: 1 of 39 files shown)


bogdanp05:
So the main change in this PR is that now the `db_session` is not defined in the functions accessing the db, as before, but in the functions calling the db access functions and it is passed as a parameter. This way you combine multiple db requests into the same session. Did I understand correctly?

FlyingBird95:
yes, this is correct. The performance is increased by reducing the amount of transactions that is made to the database. The actual number of queries is not changed.

Issue #99, opened 2018-05-02
dashboard module's templates conflict with app's templates


Harold2017:
Since the source code defines templates of dashboard module in the same folder of app's templates,
the appearance has conflict if i did not edit my base.html as the module one implies.

FlyingBird95:
Hi Harold,

thanks for noticing this. It turns out that this is intended by the Flask developers, as they formulate the problem on this site: http://flask.pocoo.org/docs/1.0/blueprints/

<img width="880" alt="screen shot 2018-05-02 at 09 10 03" src="https://user-images.githubusercontent.com/17162650/39510048-a6728852-4de8-11e8-91cc-8463ee7ce913.png">

However, this is not what we want, and therefore I changed the name of every template by prefixing it with 'fmd_'. Thus, the change of getting another clash is way lower right now.

I would like to ask you to try out our new branch: https://github.com/flask-dashboard/Flask-MonitoringDashboard/tree/issue99 and if you don't have any conflicts anymore, I'll mark this issue as closed ;)

Harold2017:
haha, we just done the same thing!
I have tried it by myself to add a prefix to all templates, and it works!

I found the same reason at here: [https://stackoverflow.com/questions/11192796/how-to-load-from-more-then-one-template-folder-for-flask-blueprint]

And this may potentially solve the problem:
`
import jinja2
custom_loader = jinja2.ChoiceLoader([
      app.jinja_loader,
      jinja2.FileSystemLoader('templates/dashboard')
])
app.jinja_loader = cusom_loader
`

This will add an alternative templates folder if the template is not found in the original one, which may partially solve this problem.

Really thank you!~

PR #100, opened 2018-05-02
renamed templates by prefixing with 'fmd_'

FlyingBird95:
this fixed issue #99


flask_monitoringdashboard/test/db/test_count.py
```diff
@@ -1,6 +1,6 @@
 """
     This file contains all unit tests that count a number of results in the database. (Corresponding to the file:
-    'dashboard/database/count.py')
+    'fmd_dashboard/database/count.py')
```
bogdanp05: The refactoring affected some of the comments and the paths don't make sense now. I'll fix this.

Issue #116, opened 2018-05-04
Make project db-independent


bogdanp05:
Remove db-specific functions such as SQLite's `strftime`.

PR #118, opened 2018-05-07
issue #116

Harold2017:
sorry, i raise a wrong pull request to master branch last time...

replace func.strftime with func.date
solve distinct query problem with MySQL

### DIFF
--- a/flask_monitoringdashboard/database/versions.py
+++ b/flask_monitoringdashboard/database/versions.py
@@ -23,6 +23,7 @@ def get_versions(db_session, end=None):
     """
     result = db_session.query(FunctionCall). \
         distinct(FunctionCall.version). \
+        group_by(FunctionCall.version). \
         filter((FunctionCall.endpoint == end) | (end is None)). \
         order_by(asc(FunctionCall.time)).all()
     return [row.version for row in result]
--- a/flask_monitoringdashboard/database/versions.py
+++ b/flask_monitoringdashboard/database/versions.py
@@ -22,8 +22,6 @@ def get_versions(db_session, end=None):
     :return: a list with the versions (as a string)
     """
     result = db_session.query(FunctionCall). \
-        distinct(FunctionCall.version). \
-        group_by(FunctionCall.version). \
         filter((FunctionCall.endpoint == end) | (end is None)). \
         order_by(asc(FunctionCall.time)).all()
-    return [row.version for row in result]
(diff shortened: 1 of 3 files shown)


flask_monitoringdashboard/database/versions.py
```diff
@@ -21,7 +21,8 @@ def get_versions(db_session, end=None):
     :param end: the versions that are used in a specific endpoint
     :return: a list with the versions (as a string)
     """
-    result = db_session.query(distinct(FunctionCall.version)). \
+    result = db_session.query(FunctionCall). \
```
bogdanp05: This method now fails a test

bogdanp05:
Hi @Harold2017 ! The first change you implemented looks fine, but the second one causes a failed test. You can do another commit to just keep the first change and then I can approve the pull request.

flask_monitoringdashboard/database/versions.py
Harold2017: Hi, i think it should be OK to pass the unittest now XD

bogdanp05:
When trying to use MySql, `get_version` method fails with:
```
sqlalchemy.exc.InternalError: (pymysql.err.InternalError) (1055, "Expression #1 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'flask_test.functionCalls.id' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by") [SQL: 'SELECT DISTINCT `functionCalls`.id AS `functionCalls_id`, `functionCalls`.endpoint AS `functionCalls_endpoint`, `functionCalls`.execution_time AS `functionCalls_execution_time`, `functionCalls`.time AS `functionCalls_time`, `functionCalls`.version AS `functionCalls_version`, `functionCalls`.group_by AS `functionCalls_group_by`, `functionCalls`.ip AS `functionCalls_ip` \nFROM `functionCalls` \nWHERE true = 1 GROUP BY `functionCalls`.version ORDER BY `functionCalls`.time ASC']
```

flask_monitoringdashboard/test/constants.py
```diff
@@ -0,0 +1,5 @@
+# Contains a number of constants that are used throughout the project
```
bogdanp05: This file should not exist.

bogdanp05:
Looks fine now!

PR #122, opened 2018-05-08
added Gitter badge to Readme

FlyingBird95:


mircealungu:
Our readme starts to look like a decorated general's uniform with all these badges!

Issue #123, opened 2018-05-09
Timezone config of plots


Harold2017:
Hi, i think adding a timezone selection may be better for displaying time data maps. Most direct but not the best way can be adding a selection field in date form. Wish you have interest in this small issue XD

mircealungu:
Wouldn't it be better to have one global dashboard setting: My Timezone which would transform all the times to the desired user timezone? Dates and times would be saved on the server in UTC and on the client side in whatever timezone the user has chosen.

FlyingBird95:
I've created a branch for this: https://github.com/flask-dashboard/Flask-MonitoringDashboard/tree/issue123
The configuration already works, but I'm not able so far to adjust the heatmap with the correspondig timezone.

But I guess that this implementation will lead to the desired solution. @Harold2017 You can also clone from this branch to continue working on!

Patrick

Harold2017:
Hi Patrick,

I just completed the database storage one in last [commit](https://github.com/flask-dashboard/Flask-MonitoringDashboard/pull/125) and make a simple trial of setting cookie [here](https://github.com/Harold2017/tests/blob/master/dashboard/test_session.pyl). And before that, i tried to store timezone info into user's config file but failed.. I found sometimes i cannot find the config file from config.file_path attr but when i use hasattr(config, 'filepath') i got true.. It is very strange so i delete that commit XD
I will try later with your solution~
Thanks!

Harold

PR #125, opened 2018-05-09
Issue#123

Harold2017:
Add a simple timezone selection field to date form


Harold2017:
issue #123

FlyingBird95:
Hi Harold,

thank you for implementing this feature. I think that it can be really useful for the Dashboard.

However, I want you to implement the following changes before I'll accept this PR:
- Add the text 'Timezone' after the Clock-icon and change the width of every column to 'col-sm-3' (including the submit-button). Then it will be on a single row. (See the figure below for what I mean)
screen shot 2018-05-09 at 10 02 42
- Set the default value to the users timezone. You can use this Stackoverflow post as an example: https://stackoverflow.com/a/17365806/9752077
- Store the users choice in the database (create a table for this in database/__init__.py). Then, the user doesn't have to change its timezone on every reload.

I'm looking forward to see more contributions from you ;)

Patrick

Harold2017:
Hi Patrick,

Thanks for your reply!
I added this timezone selection on client side but not the server side for the reason that i and my server are not in the same country XD 
I thought i should use the browser's timezone info like `new Date().getTimezoneOffset()`, but i have no idea about how to auto fill the selection form by this acquired default value in a simple way.
I will try to follow your instruction to implement server side first.
Really thank you!~

Harold

mircealungu:
the user's choice could also be stored in a cookie on the browser side... maybe less changes to the DB for this?

Harold2017:
Hi, @mircealungu 
Thanks and i will try it later~

PR #127, opened 2018-05-09
Performance

FlyingBird95:
Improved the performance of all graphs by doing the following:

In the old situation, a database request was made for every cell in a graph. 

In the new situation, the number of database requests is reduced by grouping the result of multiple queries.

A number of database functions has been changed by implementing this. Therefore, also a few tests have been updated.

### DIFF
--- a/flask_monitoringdashboard/core/plot/plots.py
+++ b/flask_monitoringdashboard/core/plot/plots.py
@@ -1,8 +1,10 @@
 """
     Contains all plots that are visible in the Dashboard
 """
+import math
 
 import plotly.graph_objs as go
+
 from flask_monitoringdashboard.core.colors import get_color
 from flask_monitoringdashboard.core.plot.util import add_default_value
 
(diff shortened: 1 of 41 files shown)


flask_monitoringdashboard/core/plot/plots.py
```diff
@@ -66,12 +68,4 @@ def get_average_bubble_size(data):
...
-    def get_max(my_list):
-        m = None
-        for item in my_list:
-            if isinstance(item, list):
-                item = get_max(item)
-            if not m or m < item:
-                m = item
-        return m
-    return get_max(data) / BUBBLE_SIZE_RATIO
+    return max([max([math.sqrt(r) for r in row]) for row in data]) / BUBBLE_SIZE_RATIO
```
bogdanp05: It's better to first find the max and then compute the sqrt, instead of computing the sqrt's of every element in the list.

PR #129, opened 2018-05-10
add timezone attr to config file and adjust time series data with corresponding timezone info

Harold2017:
Hi Patrick,

I complete your solution, please have a look~

Harold


Harold2017:
#123

FlyingBird95:
Thank you for implementing this feature.

I think that we arrived to the best solution!

Patrick

PR #138, opened 2018-05-12
added ordering for datatable

### LINKED ISSUE #137: Overview table: saving sorting criterion / or make decreasing order of hits default
1. Saving sorting criterion would be really nice to have.

2. Alternatively, simpler, and anyway better than what we have now is making the default ordering 'decreasing number of hits'. 

Personally I find myself always sorting in decreasing order of overall number of hits. I care about the most used API endpoints first. 

-- 

In any case, anything is better than alphabetical: alphabetical is good for telephone books, not for software systems. I've even wrote an entire paper about this once :) 

https://www.researchgate.net/publication/288448275_Overthrowing_the_Tyranny_of_Alphabetical_Ordering_in_Documentation_Systems

FlyingBird95:
Fixes #137

API reference: https://datatables.net/reference/api/order()

### DIFF
--- a/flask_monitoringdashboard/templates/fmd_dashboard/overview.html
+++ b/flask_monitoringdashboard/templates/fmd_dashboard/overview.html
@@ -65,3 +65,8 @@
 
 {% endblock %}
 {% endblock %}
+{% block script %}
+  <script type="text/javascript">
+    $('#dataTable').DataTable().order([[3, "desc"]]).draw();
+  </script>
+{% endblock %}


flask_monitoringdashboard/templates/fmd_dashboard/overview.html
```diff
@@ -65,3 +65,8 @@
 
 {% endblock %}
 {% endblock %}
+{% block script %}
+  <script type="text/javascript">
+    $('#dataTable').DataTable().order([[3, "desc"]]).draw();
```
bogdanp05: You are able to do the sorting in the frontend because the frontend actually has all the data from the backend, right?

FlyingBird95: that's right

PR #140, opened 2018-05-15
Add version sliders

kloostert:
Added sliders for selecting a subset of data to show in graphs.
Default slider value set to 10 (versions/users) instead of half of the available data.
Adapted graphs to show most recent data either at the top or at the right of the screen.
Made color of table rows change when hovering.


bogdanp05:
Looks nice!

PR #146, opened 2018-05-22
Schema update

bogdanp05:
Update the FunctionCalls schema


flask_monitoringdashboard/migrate.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
```
mircealungu: shouldn't the migrate have a more specific name? 
what if there's another migration in the future?

bogdanp05: True, I hadn't considered that. I'll change it to something like `migrate_v1-x-x_to_v2-x-x.py` in a future commit.

PR #156, opened 2018-06-06
updated wrapper and rules

FlyingBird95:


kloostert:
Hmm... apparently accepting PR #157 lead to the automatic merger of this PR as well.. Seems both contained the same commits.

PR #159, opened 2018-06-10
View

FlyingBird95:
Updated the view with the following JS functions:
- A JS function that converts <time>x</time> into a representative value. x must be the value in milliseconds.
- A JS function (toggleRows) that shows or hides a the children of a row, based on the fa-icon.
- automatically hide every row that has an indent >= 2.

### DIFF
--- a/flask_monitoringdashboard/static/js/custom.js
+++ b/flask_monitoringdashboard/static/js/custom.js
@@ -1,5 +1,5 @@
 (function ($) {
-    "use strict"; // Start of use strict
+    "use strict";
     // Configure tooltips for collapsed side navigation
     $('.navbar-sidenav [data-toggle="tooltip"]').tooltip({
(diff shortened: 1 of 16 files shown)


bogdanp05:
> automatically hide every row that has an indent >= 2.
This doesn't seem to work:
selection_172

flask_monitoringdashboard/static/js/custom.js
```diff
@@ -42,4 +42,46 @@
...
-})(jQuery); // End of use strict
\ No newline at end of file
+
+    // update to hide everying of hide-tag
+    $("hide").text("");
+
+    // update the duration of every html-time tag
+    $("time").text(function(i, ms){
+        // ms is a float or int
+        //ms = Math.round(parseFloat(ms) / 10) * 10;
```
bogdanp05: remove comments

PR #183, opened 2018-07-13
Issue180

bogdanp05:
Fixed compatibility issues with PostgreSQL

### DIFF
--- a/flask_monitoringdashboard/main.py
+++ b/flask_monitoringdashboard/main.py
@@ -26,6 +26,16 @@ def create_app():
         f()
         return 'Ok'
 
+    @app.route('/endpoint2')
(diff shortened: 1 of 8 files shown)


flask_monitoringdashboard/main.py
```diff
@@ -25,6 +26,16 @@ def endpoint():
         f()
         return 'Ok'
 
+    @app.route('/endpoint2')
```
mircealungu: this does not seem to be related to postgres compatibility... or?

bogdanp05: Of course not:d Just added extra endpoints to test, which I then forgot to remove

PR #190, opened 2018-09-20
Export

FlyingBird95:
- New download functionality
- Dropped support for Python 3.3 and 3.4
- Attempt to fix the grouped profiler by limiting the number of data that is analysed.

### DIFF
--- a/flask_monitoringdashboard/database/stack_line.py
+++ b/flask_monitoringdashboard/database/stack_line.py
@@ -78,9 +78,9 @@ def get_grouped_profiled_requests(db_session, endpoint_id):
     """
     t = db_session.query(distinct(StackLine.request_id).label('id')). \
         filter(Request.endpoint_id == endpoint_id). \
-        join(Request.stack_lines).limit(100).subquery('t')
+        join(Request.stack_lines).order_by(Request.id.desc()).limit(100).subquery('t')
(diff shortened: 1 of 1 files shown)


flask_monitoringdashboard/test/views/test_export_data.py
```diff
@@ -16,17 +16,17 @@ def setUp(self):
...
-    def test_download_csv(self):
+    def test_download_requests(self):
         """
             Just retrieve the content and check if nothing breaks
         """
-        test_admin_secure(self, 'download-csv')
+        test_admin_secure(self, 'download-requests')
 
-    def test_export_data(self):
+    def test_download_outliers(self):
```
bogdanp05: nice new functionality! quite useful

flask_monitoringdashboard/database/stack_line.py
```diff
@@ -78,7 +78,9 @@ def get_grouped_profiled_requests(db_session, endpoint_id):
     """
     t = db_session.query(distinct(StackLine.request_id).label('id')). \
         filter(Request.endpoint_id == endpoint_id). \
-        join(Request.stack_lines).subquery('t')
+        join(Request.stack_lines).limit(100).subquery('t')
```
bogdanp05: this will always give the data on the first 100 requests instead of the most recent 100.

PR #193, opened 2018-09-22
updated readme

FlyingBird95:


mircealungu:
problems with the questionnaire: 

for the question: "What features do you use the most in the Flask-MonitoringDashboard?"
- the answers don't match the question. 

I would suggest rephrasing the question to something like:

   "How often do you use the following features of FMD?" 

Then answers could be shortened to: 
- Don't know it
- Almost never
- Sometimes
- Often


For the question "In which country is your Flask application deployed?"
- i wonder if it wouldn't be a better question to ask in which country are *you* located? see my observations on basecamp for why this question might not be answerable. 


For "How many developers are contributing to your Flask application(s) in total?"
- see the basecamp discussion. the problem with this question might be solved by changing the tense. for zeeguu currently it's two but the total effort was more like 5 man-years. Thus i think a better question would be more: "What is the effort that went into your application until now?"

@FlyingBird95 : For the other questions, please see Basecamp!

FlyingBird95:
@mircealungu Bogdan has improved the questionnaire. Let me know if this is ok and thus can be merged?

Thanks in advance,

Patrick

PR #203, opened 2018-12-23
encode request.url into utf8

tinycolds:


FlyingBird95:
Hi,

thanks for your PR (as you see, implementing the solution works way better than posting only the issue)!

However, I still close your PR (and thus not merge), since I have made several small improvements in our development branch, so that it is better to merge that one into master.

However, I've also implemented your idea into 9c9ee7f, so it is also in our master branch from version 2.0.7 and forwards.

Again, thanks for your PR.

Patrick

PR #212, opened 2019-02-28
Update __init__.py

mircealungu:
would this be a good way to enable some extra logging?


flask_monitoringdashboard/core/config/__init__.py
```diff
@@ -99,7 +99,7 @@ def init_from(self, file=None, envvar=None, log_verbose=False):
         if envvar:
             file = os.getenv(envvar)
             if log_verbose:
-                log("Running with config from: " + (str(file}))
+                log("Running with config from: " + (str(file)))
```
bogdanp05: lol

PR #213, opened 2019-02-28
bugfix

mircealungu:
folks, is it only me or this line is required for actually reading the configuration file?

### DIFF
--- a/flask_monitoringdashboard/core/config/__init__.py
+++ b/flask_monitoringdashboard/core/config/__init__.py
@@ -106,6 +106,7 @@ class Config(object):
             return
         try:
             parser = configparser.RawConfigParser()
+            parser.read(file)
 
             # parse 'dashboard'
             self.version = parse_version(parser, 'dashboard', self.version)


mircealungu:
@bogdanp05 thanks for pulling.
is it possible that the pypi version still has the bug?

bogdanp05:
@mircealungu yes. The pypi version isn't the most recent commit on the master branch, but the one marked as a release. I'm planning to release v2.1.1 this afternoon, which will include both your bug fixes and my recent work, so the pypi version won't have the bug then.

mircealungu:
oki, cool. thanks!

bogdanp05:
@mircealungu The pypi version is now 2.1.1 and contains all the bug fixes

mircealungu:
Nice! Thanks for the heads up!

mircealungu:
I don't remember... Is the big that was making the tests freeze also fixed?

bogdanp05:
@mircealungu I don't think we did that. Are your tests freezing now?

mircealungu:
funny, they seem to not be freezing anymore :)
so you fixed even more than you thought you would
it seems!

PR #237, opened 2019-06-02
Tracking of status codes.

thejager:
Hi :)

Here's my (draft) implementation for tracking status codes. 

1. `Request` has an extra field `status_code`.
2. The status code is determined by looking at the return value of `fun()` in `measurement.py`.
3. On the endpoint page it shows the distribution of the status code:
Screenshot 2019-06-02 at 23 41 47
4. On the overview page it shows the number of requests that returned a 4xx or 5xx error:
Screenshot 2019-06-02 at 23 43 15

It's a draft because I have a few things that I need some help on:

1. The status code distribution page is on `/dashboard/endpoint/1/status_code_distribution`. How do I retrieve the value `1`? On other pages I saw it was done with `endpointService.info.id` but this value is `0` for me. Right now I temporarily solved it with a regex... :))

2. How do I make the menu to the left stay expanded on the status code page? 
Screenshot 2019-06-02 at 23 48 28

3. How would this work with migrating? A change to the DB is needed to add the `status_code` field. Should I write a migration file? Where should it go? Are there any examples from the past?

4. In the docs I read about the separation between data/logic/presentation layer. Let me know if I violated that :)

Of course any other feedback is also welcome. I will process the feedback, make sure the code is well documented and then resubmit a PR!


bogdanp05:
Hi @thejager ! I haven't looked at the PR yet, but just a quick answer for question 3: the migration files go in the [migration directory](https://github.com/flask-dashboard/Flask-MonitoringDashboard/tree/master/migration). `migrate_v1_to_v2.py` is the one more similar to what you need. Your script should be much simpler though, as in that migration we changed the db schema entirely.

flask_monitoringdashboard/controllers/requests.py
```diff
@@ -30,6 +33,21 @@ def get_num_requests_data(db_session, start_date, end_date):
     }
 
 
+def get_status_code_distribution(db_session, endpoint_id):
+    results = db_session.query(Request.status_code, func.count(Request.status_code)).filter(
+        Request.endpoint_id == endpoint_id, Request.status_code.isnot(None)).group_by(Request.status_code).all()
+
+    total_count = 0
```
FlyingBird95:
```suggestion
    total_count = sum([f for (_, f) in results])
```

flask_monitoringdashboard/controllers/requests.py
```diff
@@ -30,6 +33,21 @@ def get_num_requests_data(db_session, start_date, end_date):
...
+    for (_, frequency) in results:
+        total_count += frequency
+
+    distribution = {}
```
FlyingBird95:
```suggestion
    distribution = {status_code:frequency / total_count for (status_code, frequency) in results}
```

flask_monitoringdashboard/core/measurement.py
```diff
@@ -59,8 +59,14 @@ def add_wrapper1(endpoint, fun):
     def wrapper(*args, **kwargs):
         start_time = time.time()
         result = fun(*args, **kwargs)
+
+        if isinstance(result, tuple):
```
FlyingBird95: status_code = result[1] if isinstance(result, tuple) else 200

FlyingBird95:
Regarding your help:

1. _The status code distribution page is on `/dashboard/endpoint/1/status_code_distribution`. How do I retrieve the value `1`? On other pages I saw it was done with `endpointService.info.id` but this value is `0` for me. Right now I temporarily solved it with a regex... :))_
This is indeed using the `endpointService.info.id`, but it needs to wait a bit before it contains the right endpoint id. Therefore, you'll have to implement the logic in the `formService.setReload(function () {your code});`

2. _How do I make the menu to the left stay expanded on the status code page?_
  This is encoded the following way:
  - Every page has a unique controller. For you, this is `StatusCodeDistributionController`, which calls the menu reset function (with a unique page name).
  - The file `js/services/menu.js` contains the logic for showing/collapsing a certain menu if a page is shown. So, you'll have to change this file too.

3.  Skipping, as Bogdan already commented on this

4. _In the docs I read about the separation between data/logic/presentation layer. Let me know if I violated that :)_
It is fine ;)

flask_monitoringdashboard/controllers/requests.py
thejager: Thanks, I like this suggestion. I removed the `[]` within the `sum()` and renamed `f` to `frequency`. I think `frequency` is a bit easier to read.

flask_monitoringdashboard/controllers/requests.py
thejager: Perfect 😎

thejager:
Added a (draft) migration file. What do you think? @bogdanp05

thejager:
> Regarding your help:
> 
> 1. _The status code distribution page is on `/dashboard/endpoint/1/status_code_distribution`. How do I retrieve the value `1`? On other pages I saw it was done with `endpointService.info.id` but this value is `0` for me. Right now I temporarily solved it with a regex... :))_
>    This is indeed using the `endpointService.info.id`, but it needs to wait a bit before it contains the right endpoint id. Therefore, you'll have to implement the logic in the `formService.setReload(function () {your code});`
> 2. _How do I make the menu to the left stay expanded on the status code page?_
>    This is encoded the following way:
> 
> * Every page has a unique controller. For you, this is `StatusCodeDistributionController`, which calls the menu reset function (with a unique page name).
> * The file `js/services/menu.js` contains the logic for showing/collapsing a certain menu if a page is shown. So, you'll have to change this file too.
> 
> 1. Skipping, as Bogdan already commented on this
> 2. _In the docs I read about the separation between data/logic/presentation layer. Let me know if I violated that :)_
>    It is fine ;)

Thanks! I got it all working now!

bogdanp05:
> Added a (draft) migration file. What do you think? @bogdanp05
> 
> Not 100% sure on how to do it.

Looks good to me! Make sure to update the comments and file name.

thejager:
Wait, I forgot to add the list of non 200 status codes :) on it!

bogdanp05:
@thejager It looks fine to me, but merge it into development, rather than master.

flask_monitoringdashboard/database/request.py
```diff
@@ -33,6 +35,10 @@ def get_date_of_first_request(db_session):
     return -1
 
 
+def create_time_based_sample_criterion(start_date: datetime, end_date: datetime):
```
bogdanp05: type hinting doesn't work in python 2

thejager: Ah shit, I thought the requirement was py3. I'll fix that

flask_monitoringdashboard/controllers/requests.py
```diff
@@ -30,6 +35,67 @@ def get_num_requests_data(db_session, start_date, end_date):
...
+    """
+    return db_session \
+        .query(Request) \
+        .filter(and_(Request.endpoint_id == endpoint_id, Request.status_code.isnot(None), Request.status_code >= 400,
+                     Request.status_code <= 599, *criterion)) \
+        .all()
+
+
+def get_status_code_frequencies(db_session, endpoint_id, *criterion):
+    """
```
bogdanp05: the comment is the same as above, but the function is different.

Also, looking at the two functions (distribution and frequency), they are identical except for the past 2 lines. So the query is duplicated. Can you fix this?

flask_monitoringdashboard/static/js/services/menu.js
```diff
@@ -12,16 +12,20 @@ app.service('menuService', function ($http, endpointService) {
...
+        if (dashboardPages.includes(page)) {
             $('#collapseDashboard').collapse('show');
         } else {
             $('#collapseDashboard').collapse('hide');
         }
 
-        if (page === 'endpoint_hourly' || page === 'endpoint_user_version' || page === 'endpoint_ip' ||
-            page === 'endpoint_version' || page === 'endpoint_user' || page === 'endpoint_profiler' ||
-            page === 'endpoint_grouped_profiler' || page === 'endpoint_outlier') {
+        var endpointPages = [
```
bogdanp05: much nicer!

PR #239, opened 2019-06-12
Host performance tab

DeJonge00:
Add the host performance tab, based on the API performance tab, but grouped by hosts instead of endpoint. Includes docker-compose file with 2 FMD instance containers and a mysql database container, which should work after executing `docker-compose build` and `docker-compose up`. The containers will then be accessible from localhost ports 8080 and 8081. Migration file not included, should I make a new file or collaborate with Johan on one?


flask_monitoringdashboard/database/count.py
```diff
@@ -27,6 +27,16 @@ def count_requests(db_session, endpoint_id, *where):
     return count_rows(db_session, Request.id, Request.endpoint_id == endpoint_id, *where)
 
 
+def count_hosts(db_session):
+    """
```
bogdanp05: comments make no sense

flask_monitoringdashboard/database/host.py
```diff
@@ -0,0 +1,59 @@
...
+Contains all functions that access a Host object
+"""
+
+from flask_monitoringdashboard.database import Host, Request
+
+from sqlalchemy import asc, func, desc
+from sqlalchemy.orm.exc import NoResultFound
+
+
+def add_host(db_session, host_name: str, host_ip: str = "unknown", host_id: int=None):
```
bogdanp05: What's the role of `host_id`?

flask_monitoringdashboard/database/host.py
```diff
@@ -0,0 +1,59 @@
...
+    :return list of Host objects
+    """
+    return db_session.query(Host).order_by(asc(Host.id))
+
+
+def get_host_hits(db_session):
+    """
+    Returns all endpoint names and total hits from the database.
+    :param db_session: session for the database
+    :return list of (endpoint name, total hits) tuples
```
bogdanp05: comment seems wrong

requirements.txt
```diff
@@ -7,4 +7,6 @@ psutil                # for logging extra CPU-info
 colorhash             # for hashing a string into a color
 numpy                 # for computing median and other stats
 pytz                  # for timezone info
-tzlocal               # for figuring out the local timezone
\ No newline at end of file
+tzlocal               # for figuring out the local timezone
+pymysql               # for mysql database connections
```
bogdanp05: the last two requirements don't belong here. they can be added by the users if they do decide to use mysql or postgresql

flask_monitoringdashboard/controllers/endpoints.py
```diff
@@ -98,6 +99,22 @@ def get_api_performance(db_session, endpoints):
     } for end in db_endpoints]
 
 
+def get_host_performance(db_session, host_ids: [str], endpoint_ids: [str]):
```
FlyingBird95: nice function, good documented!!

flask_monitoringdashboard/core/config/__init__.py
```diff
@@ -41,6 +41,17 @@ def __init__(self):
         self.guest_password = ['guest_password']
         self.security_token = 'cc83733cb0af8b884ff6577086b87909'
 
+        try:
+            self.host_name = os.environ['CONTAINER_NAME']
```
FlyingBird95: use `os.environ.get(name, default)` or `if name in os.environ` instead of try-except everything

flask_monitoringdashboard/test/db/test_host.py
```diff
@@ -0,0 +1,68 @@
+"""
```
FlyingBird95: Nice, even tests ;)

FlyingBird95:
Also, I noticed that the tests fail on Python 2.7

After a small look it turned out that this is due to the typing

bogdanp05:
Is this PR going to be merged eventually?

mircealungu:
Checks fail, merge can't be done, and review comments have not been taken into account. Closing this PR for now.

PR #243, opened 2019-07-03
Fallback to default UTC timezone when no system timezone is found

nstapelbroek:
Heya 👋 

I tried setting up the dashboard on an app running in a Docker container. This caused an import failure because the timezone could not be loaded. This PR should handle the __init__ logic a bit more gracefully :) allow me to declare a timezone in the config file.

### How to test
- Setup a local environment without any resolvement methods in [unix](https://github.com/regebro/tzlocal/blob/master/tzlocal/unix.py#L39) or [windows](https://github.com/regebro/tzlocal/blob/master/tzlocal/win32.py#L89). Mounting this app in an Alpine container is a good example :)
- import the package
- Acknowledge crash with `pytz.exceptions.UnknownTimeZoneError: 'local'`
- Checkout PR
- Retry importing
- Acknowledge the import now works 🎉 

### Notes
I see the argument that you should configure your system properly. I'd  love to hear your thoughts on this in any case :)

### DIFF
--- a/flask_monitoringdashboard/core/config/__init__.py
+++ b/flask_monitoringdashboard/core/config/__init__.py
@@ -43,7 +43,11 @@ class Config(object):
 
         # visualization
         self.colors = {}
-        self.timezone = pytz.timezone(str(get_localzone()))
+        try:
+            default_timezone = pytz.timezone(str(get_localzone()))
+        except pytz.UnknownTimeZoneError:
+            default_timezone = pytz.timezone('UTC')
+        self.timezone = default_timezone
 
         # define a custom function to retrieve the session_id or username
         self.group_by = None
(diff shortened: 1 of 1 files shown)


flask_monitoringdashboard/core/config/__init__.py
```diff
@@ -43,7 +43,11 @@ def __init__(self):
 
         # visualization
         self.colors = {}
-        self.timezone = pytz.timezone(str(get_localzone()))
+        try:
+            default_timezone = pytz.timezone(str(get_localzone()))
+        except pytz.UnknownTimeZoneError:
+            default_timezone = pytz.timezone('UTC')
```
nstapelbroek: Is there a way we can log some sort of warning when this happens? I guess it makes sense to not magically handle misconfigured systems.

FlyingBird95: Hi @nstapelbroek, 

Thanks for your PR. I've slidely modified your PR and changed the base into development branche.

Thanks for your contribution, it will be accepted soon!

Patrick

PR #248, opened 2019-09-15
Fixed status code tracking for profiler level 2 and 3

thejager:


FlyingBird95:
Hi @thejager ,

thanks for the commit. Please resolve the merge conflict, then I'll accept the PR.

Kind regards,

Patrick

thejager:
Yes I saw it, will do in a second :)

thejager:
The tests seem to break, but it doesn't look like that's because of my code, is it?

bogdanp05:
I see this error in the tests: `TypeError: stop() missing 1 required positional argument: 'status_code'` You changed the `stop()` function in the profiler thread to require the status code, but this function is also used in the tests, so the tests have to be updated too.

thejager:
Alright, I was looking at this one. But the error you described makes sense. I'll fix that one and let's see what happens. Maybe it also resolves the one below.

image

bogdanp05:
Btw, do the tests pass on your local machine?

PR #259, opened 2019-10-03
Overhead

bogdanp05:
Significant performance improvements.


flask_monitoringdashboard/core/cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def display_cache():
+    """
+    Debug purposes.
+    """
+    global memory_cache
+    for k in memory_cache.keys():
+        print('%s : last=%s, avg=%f, hits=%d' % (k, memory_cache[k].last_requested,
+                                                 memory_cache[k].average_duration, memory_cache[k].hits))
```
FlyingBird95: remove

flask_monitoringdashboard/core/cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+def get_last_requested_overview():
+    """
+    Get the last requested values from the cache for the overview page.
+    """
+    global memory_cache
+    access_times = []
+    for endpoint_name, endpoint_info in memory_cache.items():
+        access_times.append((endpoint_name, endpoint_info.last_requested))
+    return access_times
```
FlyingBird95: possible rewrite as:
`return [(name, info.last_requested) for (name, info) in memory_cache.items()]`

flask_monitoringdashboard/core/cache.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def flush_cache():
+    """
+    Flushes cache changes to the db. To be called at shut down.
+    """
+    global memory_cache
+    if not memory_cache:
```
FlyingBird95: check not necessary, as an empty dict results in False

flask_monitoringdashboard/core/config/__init__.py
```diff
@@ -27,7 +27,7 @@ def __init__(self):
         self.link = 'dashboard'
         self.monitor_level = 1
         self.outlier_detection_constant = 2.5
-        self.sampling_period = 0
+        self.sampling_period = 5/1000.0
```
FlyingBird95: doesn't work in python 2.7, as it evaluates to 0

use the following import:
`from __future__ import division`

bogdanp05: 1000.0 is a float so the result will be a float, even in 2.7

flask_monitoringdashboard/core/cache.py
bogdanp05: Exactly. If the dict is empty, there's nothing to flush so the function should return

flask_monitoringdashboard/core/cache.py
```diff
@@ -0,0 +1,111 @@
...
+        self.last_requested = last_requested
+        self.average_duration = average_duration if average_duration else 0
+        self.hits = hits if hits else 0
+
+    def set_last_requested(self, last_requested):
+        with mutex:
+            self.last_requested = last_requested
+
+    def set_duration(self, duration):
+        with mutex:
```
mircealungu: do you really need the mutex here? 
i don't see any race conditions possible between set_last_requested, set_duration, and get_duration. or?

flask_monitoringdashboard/core/cache.py
```diff
@@ -0,0 +1,111 @@
...
+    It initializes the in memory cache from the db
+    """
+    global memory_cache
+    with session_scope() as db_session:
+        last_req_dict = dict(get_last_requested(db_session))
+        hits_dict = dict(get_endpoints_hits(db_session))
+        averages_dict = dict(get_endpoint_averages(db_session))
+        for rule in get_rules():
+            memory_cache[rule.endpoint] = EndpointInfo(last_requested=last_req_dict.get(rule.endpoint),
+                                                       average_duration=averages_dict.get(rule.endpoint),
```
mircealungu: which average is this one? 
is it the alltime? last 7 days? or it does not matter? 
would it make sense to add a corresponding comment in the EndpointInfo?

bogdanp05: It's alltime. It's used to determine the sleeping time of the outlier thread. But agree, will add a comment

flask_monitoringdashboard/database/endpoint.py
```diff
@@ -167,3 +169,28 @@ def get_endpoints_hits(db_session):
         join(Request). \
         group_by(Endpoint.name).\
         order_by(desc(func.count(Request.endpoint_id))).all()
+
+
+def get_avg_duration(db_session, endpoint_id):
```
mircealungu: would it make sense to limit this to the last 1K requests? 
i have an endpoint that has 100K requests... and the very
old ones are hardly much relevant... although 1K is also
quite arbitrary... :)

mircealungu:
I guess we should update also the changelog?

flask_monitoringdashboard/core/cache.py
bogdanp05: All requests made to one endpoint will update the same EndpointInfo object. If two of them come in at the same time, they'll try to update the same `hits`, `average_duration`, and `last_requested` variables, which might result in incorrect results.

flask_monitoringdashboard/core/measurement.py
```diff
@@ -73,12 +71,12 @@ def wrapper(*args, **kwargs):
...
     @wraps(fun)
     def wrapper(*args, **kwargs):
-        thread = start_profiler_thread(endpoint)
+        outlier = start_outlier_thread(endpoint)
         start_time = time.time()
         result = fun(*args, **kwargs)
         status_code = result[1] if isinstance(result, tuple) else 200
         duration = time.time() - start_time
-        thread.stop(duration, status_code)
+        outlier.stop(duration, status_code)
```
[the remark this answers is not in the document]
bogdanp05: Actually I already had them and merged them, so this PR contains the merged version:)

flask_monitoringdashboard/database/endpoint.py
bogdanp05: This is a very good point and I already addressed it in my thesis (Chapter 4.5). There was virtually no difference in response times between getting the average of 1 entry vs 100k entries.
Of course, this might differ from one rdbms to another, or there might be an arbitrary "breaking point" larger than 100k when this becomes really slow.
I think the best way to prevent this is to add a config setting for the N last requests, with default being all.

PR #260, opened 2019-10-06
Formatter

FlyingBird95:
It is quite annoying to increase the version code on every commit.

Therefore, I decided to automate this process.

I've automated this process by using python [pre-commit](https://github.com/pre-commit/pre-commit), which can be installed by:

```
cd config
pre-commit install
```

Then, I've configured the following hooks:
- [Black](https://github.com/psf/black)
- [Flake8](https://pypi.org/project/flake8)
- Increase version code: by reading the latest version from the pypi API and increasing it with 0.0.1. It is also possible to release a major or minor version, by specifying an argument to the script.

I've learned a lot about pre-commit hooks today, and there are mainly two variants: client-side and server-side. The Black and Flake8 are designed for client-side, while increasing the version is something that you want to do on the server-side. However, I couldn't manage to implement this.

Let me know if it is worth to accept this PR


FlyingBird95:
And I run black and flake8 on the entire repository, and fixed all warnings where applicable

flask_monitoringdashboard/core/cache.py
```diff
@@ -6,8 +6,12 @@
 
 from flask_monitoringdashboard.core.rules import get_rules
 from flask_monitoringdashboard.database import session_scope
-from flask_monitoringdashboard.database.endpoint import get_last_requested, get_endpoints_hits, get_endpoint_averages, \
-    update_last_requested
+from flask_monitoringdashboard.database.endpoint import (
+    get_last_requested,
+    get_endpoints_hits,
+    get_endpoint_averages,
+    update_last_requested,
```
mircealungu: i think this convention is growing on me...

flask_monitoringdashboard/test/core/test_group_by.py
```diff
@@ -2,13 +2,13 @@
 
 
 class TestGroupBy(unittest.TestCase):
-
```
mircealungu: i'm not sure I agree with Black on this one...

flask_monitoringdashboard/test/core/test_measurement.py
```diff
@@ -1,11 +1,16 @@
 import unittest
 
 from flask_monitoringdashboard.database import Endpoint
-from flask_monitoringdashboard.test.utils import set_test_environment, clear_db, add_fake_data, get_test_app, NAME
+from flask_monitoringdashboard.test.utils import (
```
mircealungu: this seems to be the main contribution of Black in 50% of the cases :)

config/.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    rev: v2.3.0
+    hooks:
+      - id: flake8
+        args: ['--max-line-length=100', '--ignore=F401, W503']
+  - repo: local
+    hooks:
+      - id: version_increase
+        name: version_increase
+        description: Increase version in constants.json
+        entry: env/bin/python config/increase_version.py
```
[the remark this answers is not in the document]
FlyingBird95: Yeah, that's not very nice indeed.
I now added a check that it only increases the version, when the branch is development.

Furthermore, the local version is only set the the remote version (from pypi) + 0.0.1. So, it will be increased at most once per merge to master.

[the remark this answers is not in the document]
mircealungu: Aaah. That's quite smart! Ignore my other comment, this makes sense!

PR #267, opened 2019-10-15
Fix status code tracking

thejager:
The original status code tracker only logged status codes properly when endpoints returned tuples. 

Now it also works when endpoints return `Response` objects, as a result of calling `jsonify()` for example.

I added some test so we can be a bit more sure about the correctness.

### Exceptions
It's a draft because exceptions are not logged yet :) working on it!


flask_monitoringdashboard/test/views/test_status_code_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+
+    app = Flask(__name__)
+
+    @app.route('/return-a-simple-string')
+    def return_a_simple_string():
+        return 'Hello, world'
+
+    @app.route('/return-a-tuple')
+    def return_a_tuple():
+        return 'Hello, world', 404
```
FlyingBird95: don't use 404 here. You can add a test for checking the 404 for an unknown endpoint.

FlyingBird95:
Two more links that are good resources for this PR:
- https://flask.palletsprojects.com/en/1.0.x/testing/
- https://flask-migrate.readthedocs.io/en/latest/

The latter one would be nice to implement ;)

flask_monitoringdashboard/test/views/test_status_code_tracking.py
thejager: But requests to non-existing endpoints aren't logged. What would you test? That there are no requests logged? That would be outside of the scope of this test.

thejager: This test covers the case where a user manually return a status code. 

And this happens. Imagine that someone requests `/posts/125` and there is no post with ID=125. In this case it would be appropriate to return status code 400.

thejager:
> Two more links that are good resources for this PR:
> 
> * https://flask.palletsprojects.com/en/1.0.x/testing/
> * https://flask-migrate.readthedocs.io/en/latest/
> 
> The latter one would be nice to implement ;)

What would I use the second link for? I don't want to make any migrations :)

flask_monitoringdashboard/test/views/test_status_code_tracking.py
FlyingBird95: ah, oke. I forgot that we don't monitor 404 endpoints ;) Now it makes sense

FlyingBird95:
> What would I use the second link for? I don't want to make any migrations :)

Nevermind. I thought you added something to the DB

thejager:
> do you still return the result if the exception is raised?

No, but if an exception was raised in a route handler, there is no result to be returned :)

PR #282, opened 2019-12-05
[parser.py] Avoid to raise exception in case of corrupt git folder

alessiosavi:
Avoid to raise exception in case of corrupt git folder


flask_monitoringdashboard/core/config/parser.py
```diff
@@ -23,9 +23,14 @@ def parse_version(parser, header, version):
...
-            # cut version to at most 6 chars
-            return version[:6]
+            version_file = os.path.join(git , git_file)
+            if os.path.exists(version):
+                version = open(version).read()
+                # cut version to at most 6 chars
+                return version[:6]
+            else:
+                # Return dummy version in case of no git version file found
+                return "v0.0.1"
```
FlyingBird95: Please replace `return "v0.0.1"` with `return version`.

Then I'll approve it.

Thanks for your PR,

Patrick

flask_monitoringdashboard/core/config/parser.py
```diff
@@ -23,9 +23,14 @@ def parse_version(parser, header, version):
             # The file is specified by: 'ref: <location>'
             git_file = (open(os.path.join(git, 'HEAD')).read().rsplit(': ', 1)[1]).rstrip()
             # read the git-version
-            version = open(git + '/' + git_file).read()
-            # cut version to at most 6 chars
-            return version[:6]
+            version_file = os.path.join(git , git_file)
```
bogdanp05: I like this, we should avoid manually concatenating paths

alessiosavi:
Hi Team,

Of course, `os.path.join` will take care about joining the path, avoiding headache :D

Why i raise this pull request:
image

flask_monitoringdashboard/core/config/parser.py
alessiosavi: Done Sir!

flask_monitoringdashboard/core/config/parser.py
alessiosavi: Of course, os.path.join will take care about joining the path, avoiding headache :D

flask_monitoringdashboard/core/config/parser.py
```diff
@@ -23,9 +23,14 @@ def parse_version(parser, header, version):
...
+            if os.path.exists(version):
+                version = open(version).read()
+                # cut version to at most 6 chars
+                return version[:6]
+            else:
+                # Return "dummy" version in case of no git version file found
```
mircealungu: the comment doers not make much sense anymore if we return `version` now. or? 
actually, that's a good question: what's version in this case?

FlyingBird95: `version` is set from `APP_VERSION`. Or, in case this environment-variable, doesn't exist, it is `config.verson`.

alessiosavi: It use `1.0.` as pointed by 
`core/config/__init__.py:26:        self.version = '1.0'`

alessiosavi:
There was a mistake in this pull request.

I've fixed in 248de77671ba185d8c51a974e4bc0108ef0e79f9

PR #286, opened 2019-12-11
fix request set down

rileyhun:
I think the Outlier table is trying to update before the Request table is updated. The Outlier table has a foreign key that's linked to a primary key in the Request table, so by the time the Outlier table is updated, the corresponding record in the Request table does not exist. 

Should we commit our changes to the Request table right after the flush?

### DIFF
--- a/flask_monitoringdashboard/database/request.py
+++ b/flask_monitoringdashboard/database/request.py
@@ -57,6 +57,7 @@ def add_request(db_session, duration, endpoint_id, ip, group_by, status_code):
     )
     db_session.add(request)
     db_session.flush()
+    db_session.commit()
     return request.id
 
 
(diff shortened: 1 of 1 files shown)


flask_monitoringdashboard/database/request.py
```diff
@@ -57,6 +57,7 @@ def add_request(db_session, duration, endpoint_id, ip, group_by, status_code):
     )
     db_session.add(request)
     db_session.flush()
+    db_session.commit()
```
bogdanp05: Based on the error message, your analysis seems correct. We actually added at some point `db_session.flush()` to fix precisely this problem, but apparently it wasn't enough. 

Looking at this [post](https://medium.com/@oba2311/sqlalchemy-whats-the-difference-between-a-flush-and-commit-baec6c2410a9) and this [SO answer](https://stackoverflow.com/a/4202016), it appears that `commit()` always calls `flush()`, so `db_session.flush()` should be removed since it's redundant.

PR #293, opened 2020-02-03
Api requests status

jangster3794:
A user can now click on a button to display success calls and failed calls
Status Code range:
success: 100 to 400
failure: 400 to 599

image


flask_monitoringdashboard/database/stack_line.py
```diff
@@ -42,12 +42,17 @@ def get_profiled_requests(db_session, endpoint_id, offset, per_page):
     :return: A list with tuples. Each tuple consists first of a Request-object, and the second part
     of the tuple is a list of StackLine-objects.
     """
+    start_code, end_code = 0,600
+    if response_status == "failed":
+        start_code=400
+    elif response_status == "success":
+        end_code=400
```
mircealungu: this includes 400 for success.

flask_monitoringdashboard/static/elements/pagination.html
```diff
@@ -2,6 +2,42 @@
...
+                <h6>|</h6>
+            </div>
+            <div>
+                <a href="#" ng-click="pagination.updateStatusCode('success')">Success</a>
+            </div>
+            <div>
+                <h6>|</h6>
+            </div>
+            <div>
+                <a href="#" ng-click="pagination.updateStatusCode('failed')">Failed</a>
```
mircealungu: either Success | Fail or Succeed*ed* | Fail*ed*

flask_monitoringdashboard/views/profiler.py
```diff
@@ -19,8 +19,10 @@ def num_profiled(endpoint_id):
 @blueprint.route('/api/profiler_table/<endpoint_id>/<offset>/<per_page>')
 @secure
 def profiler_table(endpoint_id, offset, per_page):
+    sort_by = request.args.get('sortby', 'latest')
+    response_status = request.args.get('response_status', 'all').lower()
     with session_scope() as db_session:
-        return jsonify(get_profiler_table(db_session, endpoint_id, offset, per_page))
+        return jsonify(get_profiler_table(db_session, endpoint_id, offset, per_page, sort_by, response_status))
```
mircealungu: In sync with the other param being called `sort_by` maybe `response_status` should be called `filter_by`? Or maybe `filter_by_status` ?

flask_monitoringdashboard/static/elements/pagination.html
jangster3794: The way Buttons are there is at least one of them will be selected so that's why All is there to filter all irrespective of the response code. So I would suggest not changing this.

I have made other changes already, please do review them.

flask_monitoringdashboard/views/profiler.py
jangster3794: @mircealungu @FlyingBird95 Can you please review the changes and help me out here? Thanks.

flask_monitoringdashboard/static/elements/pagination.html
mircealungu: Perfectly find that All is there. 
I was referring to the *form* of Sucess / Fail. 
It should match: if the first is Success then the second should be Fail. 
Or Succeeded / Failed. But if you mix them it looks bad.

flask_monitoringdashboard/views/profiler.py
mircealungu: Dear @jangster3794, 

Thanks for the reminder,  and apologies for the slowness of my response. 

I looked again at the behavior of the system with this PR and I think it's still buggy. 
What do you think? Let me higlight the steps to replicate: 

1. Before filtering based on response code I see this: 

Screen Shot 2020-03-04 at 3 00 05 PM

2. After filtering only the Failed I see this: 
- note that although there's only one request the info at the top mentions stil 24K and I still have the choice of selecting other pages

Screen Shot 2020-03-04 at 3 00 27 PM

3. And obviously, when I select page 2 I get no request listed, but twice the navigation pannels...

image

jangster3794: Hi,
Sorry for the late reply.

Thanks for highlighting these bugs.
I am going to start work on these bugs and update this PR in a couple of days.

bogdanp05:
Is this still active?

mircealungu:
this is dormant for too long. 
i'll close it for now but it can be reopened if progress is made.

PR #321, opened 2020-04-05
Improvements to reports

### LINKED ISSUE #318: Significancy threshold is too high for for mean latency reporting
Guys,

currently mean latency differences are only reported if the difference is at least 30%. Motivation was: differences less than 30% aren't really interesting. On second thought, I think even a change of a few percent can matter to some applications.

I suggest we lower this threshold or remove it all together.

What do you think?

### LINKED ISSUE #319: Results with very little evidence are also included in reports
image

This example claims a 53% decrease in latency, but for these two intervals there are only two measurements for this endpoint. Obviously, ridiculous to include a result like this, given such little evidence.

I have two simple suggestions:
- Include all results, but mention the size of the samples used. Add somewhere: "Based on two samples with 1 measurement and 2 measurements". This way people can make their own judgement about the validity of this result.
- Exclude results that compared small samples -> Either sample needs to have at least 30 measurements for example.

These are two simple options that only take a few lines at most to implement. There are probably more sophisticated ways, but I'll research those during my thesis project :)

thejager:
On evaluation of the current reports, I discovered that the statistical validity of the reports were questionable at best. This PR fixes the following issues:

- Latency measurements follow a distribution with a long tail to the right. Means don't say much about data like this. Means have been replaced by medians.
- Performance differences had to be at least 30% in the previous version even though a few percent difference can already be significant for some projects. Threshold removed altogether.
- In the previous version, a randomised permutation test was used to determine whether the two samples came from two different distributions. Could not very find much support for this method, so I switched to a more commonly used non-parametric test that's appropriate for medians.
- Moved the label "Execution Time (ms)" to the y axis, where it is supposed to be.
- Removed some unused code (in the reporting module).
- Fixed formatting in some files. Formatter didn't seem to have ran in the previous version of the code.
- Renamed "Interesting" to "Significant" in the report.

Fixes #318 and fixes #319 

For a more detailed motivation, see the report that was written for my research internship :)


FlyingBird95:
@mircealungu 
if you like this PR, feel free to accept it ;)

mircealungu:
Ehi @FlyingBird95 and @thejager !

I'll look at it tomorrow after coffee :)

flask_monitoringdashboard/core/reporting/questions/median_latency.py
```diff
@@ -0,0 +1,95 @@
...
+from flask_monitoringdashboard.database import session_scope
+
+from flask_monitoringdashboard.database.request import get_latencies_sample
+
+
+class MedianLatencyReportAnswer(ReportAnswer):
+    def __init__(
+        self,
+        is_significant,
+        comparison_interval_latencies_sample=None,
```
mircealungu, line 17: should we maybe start adding type annotations?

flask_monitoringdashboard/core/reporting/questions/median_latency.py
```diff
@@ -0,0 +1,95 @@
...
+        compared_to_interval_latencies_sample=None,
+        percentual_diff=None,
+        comparison_interval_median=None,
+        compared_to_interval_median=None,
+    ):
+        super().__init__('MEDIAN_LATENCY')
+
+        self._is_significant = is_significant
+        self._comparison_interval_latencies_sample = comparison_interval_latencies_sample
+        self._compared_to_interval_latencies_sample = compared_to_interval_latencies_sample
```
mircealungu, line 27: would the term *baseline* make sense here? if the meaning is the same reading it might be faster because it's a completely different word.

flask_monitoringdashboard/core/reporting/questions/median_latency.py
```diff
@@ -0,0 +1,95 @@
...
+        self._percentual_diff = percentual_diff
+
+        self._compared_to_interval_median = compared_to_interval_median
+        self._comparison_interval_median = comparison_interval_median
+
+    def meta(self):
```
mircealungu, line 33: a brief comment on what's meta would be nice.
here or in the superclass.

flask_monitoringdashboard/core/reporting/questions/status_code_distribution.py
```diff
@@ -27,53 +28,69 @@ def frequency_to_percentage(freq, total):
...
+                        compared_to_interval=compared_to_interval_percentage,
+                        percentage_diff=percentage_diff,
+                    )
+                )
 
                 max_absolute_diff = max(max_absolute_diff, percentage_diff)
 
-            return StatusCodeDistributionAnswer(is_significant=max_absolute_diff > 3, percentages=percentages)
+            return StatusCodeDistributionReportAnswer(
+                is_significant=max_absolute_diff > 3, percentages=percentages
```
mircealungu, line 95: magic constant... give symbolic name maybe

flask_monitoringdashboard/core/reporting/questions/median_latency.py
thejager: I would love that, but we still support Python 2.7, right? I don't think that is supported in Python 2.7.

In that case, is it an idea to drop Python 2.7 support soon?

flask_monitoringdashboard/core/reporting/questions/median_latency.py
thejager: Thanks, that's much better! I couldn't come up with a good name...

flask_monitoringdashboard/core/reporting/questions/median_latency.py
mircealungu: ah, you're right; forget about it for now :)

thejager:
Just processed your feedback @mircealungu 

I think this cleaned it up a lot! Especially the variable renaming. It's now much easier to distinguish the two from each other. 

Let me know what you think

flask_monitoringdashboard/core/reporting/questions/report_question.py
```diff
@@ -1,8 +1,7 @@
-
 from abc import ABCMeta, abstractmethod
 
 
-class Answer:
+class ReportAnswer:
```
mircealungu, line 4: i like this name more :)

flask_monitoringdashboard/core/reporting/questions/report_question.py
```diff
@@ -14,13 +13,14 @@ def is_significant(self):
 
     @abstractmethod
     def meta(self):
+        """
+        Should return a `dict` that contains any additional information required on the
+        frontend. This will be included in the response.
```
mircealungu, line 18: cool! dank u wel!

flask_monitoringdashboard/main.py
```diff
@@ -16,8 +16,9 @@
 
 dashboard.config.version = '3.2'
 dashboard.config.group_by = '2'
-dashboard.config.database_name = 'sqlite:///data.db'
-# dashboard.config.database_name = 'mysql+pymysql://user:password@localhost:3306/db1'
+dashboard.config.database_name = 'mysql://root:my-secret-pw@127.0.0.1:3306/fmd'
```
mircealungu, line 19: is this a better connection string than the other ones?

mircealungu: this is the module for quick testing of the FMD right? 
for testing probably it'd be simpler with sqlite: you don't have to create a DB in mysql...or?

thejager: Argghh, I am using a mysql database for development and I keep accidentally pushing this file. I should make a copy of this file. It's definitely supposed to be the sqlite connection string! Good catch :)

Issue #311, opened 2020-03-26
How to use the flask-monitoringdashboard offline?


MustNull:

MustNull:
> Hi @MustNull! What exactly do you mean by using it offline? We have a [documentation page ](https://flask-monitoringdashboard.readthedocs.io/en/latest/installation.html#installing-the-flask-monitoringdashboard-package)detailing how to get started, hope it helps:)
Read this code, there are some js file need network. But my flask server cannot access to internet.
https://github.com/flask-dashboard/Flask-MonitoringDashboard/blob/0404b05b9a8f1917796e0f314a77a53a754a0b15/flask_monitoringdashboard/static/js/controllers/overview.js#L26

thejager:
I don't see what exactly the problem is. Surely the request will fail, but the FMD still works, right?

Can you explain me why exactly this is a problem? It would be a possibility to add an option to enable offline mode in the settings, but I don't see the added value yet.

MustNull:
Actually i cannot see anything i want if i deploy this application in offline environment.
The web site show nothing except login table.
I cannot see anything even if i login in.

GalDayan:
Hey,

I also need this offline support.

Is there any way to make it work offline?

mircealungu:
@GalDayan - we don't have much engineering power at the moment. 
Do you want to look into it and send a PR with a fix?

GalDayan:
> @GalDayan - we don't have much engineering power at the moment.
> Do you want to look into it and send a PR with a fix?

I will prioritize this task and I will update :)
Thanks for replying.

GalDayan:
> @GalDayan - we don't have much engineering power at the moment.
> Do you want to look into it and send a PR with a fix?

It seems like it already merged in #360, is it?

PR #322, opened 2020-04-16
Modernise the frontend

thejager:
Hey guys,

I was working a bit on my webpack skills and thought FMD could really make use of this. So I rewrote the frontend so it makes use of one javascript file that imports all its required files. Webpack bundles this file into 1 minified file

Why?

- Shortened load times. Because of tree shaking and minification the total amount of assets is now almost half of what it was.
- No issues with order of imports.
- Can use modern JS features without having to worry about compatibility with old browsers. Babel takes care of this.
- Proper package management. Easy updates, no risk of installing incompatible versions of packages.
- Sass!


FlyingBird95:
Hey Johan,

Awesome to see this pull request! I had this idea in mind for a long time, but I didn't have the time/skills to do this. I think that this is definitely the way to go with the frontend. But I have a few considerations:
- How are you planning to do the build, since we're shipping a Python package, we cannot guarantee that people have `npm` installed. Also, if people have to run `npm i && npm build` first, than we don't have the easy deploy and go as we have now (this is one of our selling points). I have several solutions for this, but I'm not sure what is the best option:

  - We can keep the build into the `static`-folder. Our frontend build contains dependencies, which results in big 'build'. Also, our Github repo will look like a JS project then. In general it is never a good idea to keep compiled files into our repo. However, this solution is the simplest.
  - We can put the frontend code into a separate repo, and keep the build in this seperate repo. This doesn't solve all points as described in my previous solution, but we then have a nice separation between frontend (JS) and backend (Python). We can use the github servers as our hosting solution (e.g. get the JS code from https://raw.githubusercontent.com/flask-dashboard/Flask-frontend/master/dist/js/app.js). However, Github is not made for being a hosting server.
  - The last option that I have in mind (and certainly the most professional option) is an extension of my previous solution, but configure an automatic build to an online hosting option. I recently discovered Google Firebase, which can definitely do the job. We don't have to keep our compiled files in the repo anymore, and a push to master is automatically deployed on Google Firebase. I already started experimenting with it a week ago, and managed to deploy some source files to https://flask-monitoringdashboard.web.app, which are then build automatically.

So, as you can read from my options above, I have strong preference for the last one. However, I'm not sure yet how to do it with versioning (I know that Google Firebase supports this, but I'm not sure how we can configure this). Also, I'm not sure about the pricing, but I think that it is sufficient for us (see here more: https://firebase.google.com/pricing). The decision that we have to make is how 'professional' we would like to have our FMD? More professional means the more time we have to put into it ;)

I would love the hear your opinion about this, and I would like to emphasize that I really like your effort.

Have a nice weekend,

Patrick

p.s. I would also like to have an opinion from @mircealungu @vandriko @bogdanp05

thejager:
Hi Patrick,

thanks for your comment! :)

I think my preference actually goes out to 1.

> our Github repo will look like a JS project then

Because there is 1 `app.js` file? I don't see how that would make it seem like a JS project. Or do you mean the code distribution thing that shows up on the github page? I did some googling and it seems like you can add a `.gitattributes` file to ignore certain files from showing up in those statistics.
image

> In general it is never a good idea to keep compiled files into our repo.

You mean specifically in the FMD repo? Because generally speaking most libraries do this. They ship with the original source files and the transpiled/minified build. If you look at the Bootstrap repo for example it has a `js/` and a `/dist` folder for respectively the source and the build.

-- 
It actually comes with an advantage the other two options don't have: Because all dependencies are compiled into the app.js/app.css, you don't need requests to external CDNs. I am not sure how many people care about this, but I think it can be valuable to people running some kind of intranet for example that don't have access to internet.

Definitely not a fan of the second option you suggested. One project, one repo :)

FlyingBird95:
Hey Johan,

> Or do you mean the code distribution thing that shows up on the github page?

Yes, that's what I meant. The `.gitattributes` seems like a nice solution.

I think you're right about the `/dist` files. I suggest to put all frontend source code into `/frontend` so that the only the build code is shipped in the Python package. We then have to configure webpack, such that the build code ends up in our static folder `/flask_monitoringdashboard/static`.

Actually, I like the idea of a single `.js`-file with all dependencies now (so we don't have to request CDN urls anymore). This will also (partially) solve #311 

I'm very much looking forward into the result!

Patrick

mircealungu:
Hi guys!

By the time I join the conversation the discussion is solved already :)
I wish more student projects would be like this one :)

I like the idea of separating the JS in a /frontend folder away from the Python code!

Looking forward to seeing how it looks!

thejager:
Can anyone give this PR a shot?

In the `/frontend` directory you can run `npm run dev` to compile the assets. This will place the compiled JS and CSS in the `static/` directory

bogdanp05:
Hi @thejager! I tried to run `npm run dev` but I get
```
> frontend@1.0.0 dev /Users/bogdan/Flask-MonitoringDashboard/flask_monitoringdashboard/frontend
> webpack --watch

sh: webpack: command not found
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn
npm ERR! frontend@1.0.0 dev: `webpack --watch`
npm ERR! spawn ENOENT
npm ERR!
npm ERR! Failed at the frontend@1.0.0 dev script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/bogdan/.npm/_logs/2020-05-19T23_39_20_085Z-debug.log
```

I'm pretty bad at frontend stuff so maybe I'm missing something obvious. My npm version is 6.9.0

bogdanp05:
I figured it out eventually. I had to do `npm install webpack` to install Webpack. Then `npm run dev` still failed with 
```
ERROR in ./sass/app.scss
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
Error: Node Sass does not yet support your current environment: OS X 64-bit with Unsupported runtime (83)
For more information on which environments are supported please see:
https://github.com/sass/node-sass/releases/tag/v4.13.1
    at module.exports (/Users/bogdan/Flask-MonitoringDashboard/flask_monitoringdashboard/frontend/node_modules/node-sass/lib/binding.js:13:13)
```

But theeen, I found [this page](https://medium.com/@proustibat/how-to-fix-error-node-sass-does-not-yet-support-your-current-environment-os-x-64-bit-with-c1b3298e4af0), so I ran
```
npm rebuild node-sass
```
and it took a few minutes, but after that `npm run dev` worked.

flask_monitoringdashboard/frontend/js/filters.js
```diff
@@ -0,0 +1,124 @@
...
+    app.filter('duration_ms', function () {
+        return function (time) {
+            return Math.round(parseFloat(time) * 10) / 10;
+        }
+    });
+
+    app.filter('duration', function () {
+        return function (time) {
+            // ms is a float or int
+            //ms = Math.round(parseFloat(ms) / 10) * 10;
```
bogdanp05, line 12: remove commented code

PR #325, opened 2020-04-27
Update README.md

mircealungu:
badges seem to need an empty line below the <span>

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ A dashboard for automatic monitoring of <a href="http://flask.pocoo.org/" target
 </p>
 
 <span align="center">
+    
 ![Build Status](https://github.com/flask-dashboard/Flask-MonitoringDashboard/workflows/Python%20Testing/badge.svg)
 [![Documentation Status](https://readthedocs.org/projects/flask-monitoringdashboard/badge/?version=latest)](http://flask-monitoringdashboard.readthedocs.io/en/latest/?badge=latest)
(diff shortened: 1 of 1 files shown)


FlyingBird95:
Hey Mircea,

thanks for spotting this issue. The fix looks good!

I see that the build is failing. The logs show that this is due to `scipy`. In [the documentation](https://www.scipy.org/scipylib/faq.html#do-numpy-and-scipytill-support-python-2-7) I've found that `scipy` supports until 1.2.x on Python 2.7. Maybe we should update our `requirements.txt` and replace `scipy==1.4.1` to only `scipy` (option 1), or change into `scipy==1.2.2` (option 2). What option do you think is the best?

Kind regards,

Patrick

mircealungu:
I think it's probably safe to pin the dependency to `scipy==1.2.2` for this year, or however long we still support 2.7. 

How would the problem be fixed with option 1?

FlyingBird95:
I think that users of our FMD have more freedom with option 2. Suppose that you already use `scipy` in your project, and the version that you're using is `1.3.x`. Then it won't be possible to use the FMD with `scipy==1.2.2`, due to different versions. Up to you to choose, I'm fine with both

Patrick

mircealungu:
but then the build will still fail. or?

FlyingBird95:
Short answer: one way to find out ;)

Long answer: I don't think so, as it will install the latest possible version for the python version. But this will result into different versions of `scipy` used in 2.7 and 3.X. Do we want this?

mircealungu:
cool. 
let’s find out then :)
m.

PR #334, opened 2020-05-12
Create codecov.yml

FlyingBird95:
This PR will automatically fail all PR's that drop the codecoverage percentage by more than 1%.

Currently this is configured as 0% (so a decrease in code coverage, means that it is not allowed to pass).

I think that 0% is a bit too strict, so let's use a value of 1%.


FlyingBird95:
Ironically, this doesn't work. Let's try something else ;)
Screenshot 2020-05-12 at 16 35 42

mircealungu:
1% is fair enough :)

PR #335, opened 2020-05-16
Improve testing and update some code

FlyingBird95:
Hey all,

since this is a huge pull request, I would like to have a review from @mircealungu, @bogdanp05, and @thejager. Please take a moment to review all 121 changed files (yes, I already told you that it is huge 😉), and leave comments about stuff that is not clear. The most notable changes are:

* Testing is now performed using `pytest` instead of `unittest`. More on this below.
* I renamed the variable `db_session` into just `session`. This is easier to read.
* I updated a lot of docstrings. We tend to do this in the form of [1], but it can be way shorter in the form of [2].

  [1]:
  ```
  """
      This is our comment.
  """
  ```
  [2]:
  ```
  """This is our comment."""
  ```
* I updated `python-publish.yml` so that we also deploy our development version to pypi. This makes it easier to deploy the development version to an existing application, which leads to finding bugs more easily (and it is better to find them on the development branch of course). I haven't actually tested whether this functionality works, so let's see.
* I renamed files to use the python convention, i.e. `StringHash.py` becomes `string_hash.py`. from camelCase to snake_case 😅.
* I updated the comments in `flask_monitoringdashboard/database/__init__.py` below the variables, which is the python convention.

**pytest**
Now more details about `pytest`. It works with fixtures, which might be defined like this:

```
@pytest.fixture
def x():
    return 42
```

We can use fixtures as arguments of our tests. Let's say that we have a function `f`, and we want to test the outcome of this function:

```
def f(n):
    return abs(n) * 2 + 5


def test_f(x):
    assert f(x) == 89
```

This is a silly test, but we can easily test for other values of `x`, like this:

```
import pytest

@pytest.mark.parametrize('x', [5])
def test_f_2(x):
    assert f(x) == 15
```

Since, the `parameterize` takes a list of arguments, we can also create 2 tests with the same amount of lines, like this:

```
import pytest

@pytest.mark.parametrize('x', [5, -5])
def test_f_2(x):
    assert f(x) == 15
```

So, this is a minimal introduction to pytest. For more info, take a look [at this site](https://docs.pytest.org/en/latest/).

You might wonder how this is usefull for the FMD. That's where the `pytest_factoryboy` comes in. This library turns a database model into a fixture. All these fixtures are defined in `tests/fixtures/models.py`. 

The bottom of this file registers them as actual features. A small snapshot:
```
register(EndpointFactory, 'endpoint')
register(RequestFactory, 'request_1')
register(RequestFactory, 'request_2')
```

And the nice thing about this is that they are connected. If we want to have an endpoint for testing without any requests, we can simply do:

```
def test_endpoint(endpoint):
    assert endpoint.requests == []
```

Or with one request:
```
def test_endpoint_with_request(request_1, endpoint):
    assert endpoint.requests == [request_1]
    assert request_1.endpoint == endpoint
```

And the same thing for two requests. A last function to introduce to you is `usefixtures`. You can use this if you need for example two requests for the endpoint that you're testing, but don't actually use the arguments in the code. It looks like this:

```
import pytest


@pytest.mark.usefixtures('request_1', 'request_2')
def test_endpoint_with_two_requests(endpoint):
    assert len(endpoint.requests) == 2
```

All this stuff above are the building block of using pytest. Using this, I have rewritten all tests completely. It took me a couple of weeks, but it was a nice exercise. I really appreciate it if you can find a moment to review my PR. If you have any suggestions or improvements, please let me know!

@thejager : I also added a test about the reporting, but it is only testing the flow for which nothing is significant. Do you want to add a test that tests the significant part?


FlyingBird95:
One comment that I forgot to make is that we don't test the interaction of our frontend with the backend (JS code that makes the requests and Python that handles it). The reason for leaving this out of the PR is that I'm not super familiar with this. Also, I don't really know what's the best approach here. Some plain JS tests, or full functional tests using Selenium?

@thejager What do you suggest?

tests/api/test_endpoint.py
```diff
@@ -0,0 +1,222 @@
...
+from flask_monitoringdashboard.database import Endpoint, Request
+
+
+@pytest.mark.parametrize('endpoint__monitor_level', [3])
+def test_overview(dashboard_user, request_1, request_2, endpoint, session):
+    response = dashboard_user.get('dashboard/api/overview')
+    assert response.status_code == 200
+
+    [data] = [row for row in response.json if row['id'] == endpoint.id]
+    print(data)
```
bogdanp05, line 12: is this supposed to be here?

flask_monitoringdashboard/controllers/__init__.py
```diff
@@ -1,3 +1 @@
-"""
-This directory is used for implementing the logic between the database and the API
-"""
+"""This directory is used for implementing the logic between the database and the API."""
```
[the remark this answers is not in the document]
FlyingBird95: This is indeed something that I'm planning to do in another PR. Note that we already have a section in Github Actions with `flake8`. This now uses the `--exit-zero` argument, which means that we don't fail the test suite when any constraint (such as max 100 chars per line) is violated.

flask_monitoringdashboard/core/config/__init__.py
```diff
@@ -42,7 +40,7 @@ def __init__(self):
         # authentication
         self.username = 'admin'
         self.password = 'admin'
-        self.guest_username = ['guest']
+        self.guest_username = 'guest'
```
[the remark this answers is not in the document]
FlyingBird95: it seems like this is a bug in our FMD. Can @mircealungu check if he can login using a guest credentials?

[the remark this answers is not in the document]
mircealungu: Good question. Indeed, I can't login with guest/guest

.github/workflows/python-publish.yml
```diff
@@ -1,8 +1,10 @@
 name: Upload Python Package
 
 on:
-  release:
-    types: [created]
+  push:
+    branches:
+      - master
+      - development
```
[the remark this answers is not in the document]
mircealungu: I'm also not sure that I see the distinction between `development` and 'master` in this case.

flask_monitoringdashboard/database/__init__.py
```diff
@@ -27,115 +27,168 @@
...
-    # the HTTP status code of the request
+    """IP address of the requester."""
+
     status_code = Column(Integer, nullable=True)
+    """HTTP status code of the request."""
 
-    endpoint = relationship(Endpoint)
-    stack_lines = relationship('StackLine', back_populates='request')
-    outlier = relationship('Outlier', uselist=False, back_populates='request')
+    outlier = relationship("Outlier", uselist=False, back_populates='request')
```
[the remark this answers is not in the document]
FlyingBird95: The StackLine table looks like this:
```
class StackLine(Base):
    ...
    request = relationship(Request, backref="stack_lines")
```

Since it is using `backref` (instead of `back_populates`), it adds the property `stack_lines` to Request.

flask_monitoringdashboard/database/request.py
```diff
@@ -8,39 +8,28 @@
...
-            .limit(sample_size) \
-            .all()
+    query = (
+        session.query(Request.duration)
+        .filter(Request.endpoint_id == endpoint_id, *criterion)
+        .limit(sample_size)
+    )
 
-        durations = [item.duration for item in items]
+    # dialect = session.bind.dialect.name
```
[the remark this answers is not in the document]
FlyingBird95: done

tests/api/test_endpoint.py
FlyingBird95: good spot!

flask_monitoringdashboard/__init__.py
```diff
@@ -22,21 +22,21 @@
 config = Config()
 
 
-# get current location of the project
 def loc():
+    """Get the current location of the project."""
```
mircealungu, line 26: nice to have the """ comments!

flask_monitoringdashboard/core/config/__init__.py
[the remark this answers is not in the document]
mircealungu: this could be an opportunity to use a type annotation for a class member variable!

flask_monitoringdashboard/database/__init__.py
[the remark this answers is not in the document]
bogdanp05: You're right. I didn't see you updated the `request` relationship in `StackLine`

.github/workflows/python-publish.yml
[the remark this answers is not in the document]
FlyingBird95: My initial idea was to have a version like `3.2.0` and `3.2.0-dev` as the development on PyPi. After doing some research (I found this post), I think that it is indeed something that needs to be reverted: https://stackoverflow.com/questions/16692799/development-version-on-pypi.

I will revert this in a later case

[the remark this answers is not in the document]
mircealungu: Ah! Now I understand what you mean! 
Might be a cool idea after all :)
People who want the bleeding edge go with `-dev`, the rest stay with the master.
Do you know other Python libraries who do this? 
I see that in your SO post the two examples (django and lxml) don't use it, but there might be good examples of usage?

flask_monitoringdashboard/main.py
```diff
@@ -48,8 +49,8 @@ def endpoint():
     # if session_scope is imported at the top of the file, the database config won't take effect
     from flask_monitoringdashboard.database import session_scope
 
-    with session_scope() as db_session:
-        print(db_session.bind.dialect.name)
+    with session_scope() as session:
+        print(session.bind.dialect.name)
 
     print("Hello, world")
```
mircealungu, line 55: not sure this is useful for anything.

migration/migrate_v1_to_v2.py
```diff
@@ -49,8 +49,8 @@ def get_connection(db_url):
 def session_scope():
     """
     When accessing the database, use the following syntax:
```
mircealungu, line 51: wasn't the convention that comments start on the same line with """?

tests/unit/database/test_request.py
```diff
@@ -0,0 +1,66 @@
...
+            ip='127.0.0.1',
+            group_by=None,
+            status_code=200,
+        )
+        assert count_requests(session, endpoint.id) == num_requests + 1
+
+
+@pytest.mark.parametrize('request_1__time_requested', [datetime(2020, 2, 3)])
+def test_get_versions(session, request_1):
+    """Test whether the function returns the right values."""
```
mircealungu, line 44: this comment is copy-pasted across the following three tests.
do we really need it?

tests/fixtures/dashboard.py
```diff
@@ -0,0 +1,63 @@
...
+
+import flask_monitoringdashboard
+
+
+@pytest.fixture
+def config(username='admin', password='admin', guest_username=None, guest_password=None, colors=None, group_by=None):
+    flask_monitoringdashboard.config.username = username
+    flask_monitoringdashboard.config.password = password
+    flask_monitoringdashboard.config.guest_username = guest_username or 'guest'
+    flask_monitoringdashboard.config.guest_password = guest_password or ['guest_password']
```
mircealungu, line 13: would it make sense to rename this into `guest_passwords` so it's clear that it's a list?

tests/unit/core/config/test_config.py
```diff
@@ -0,0 +1,49 @@
...
+    parser.set(section, 'literal2', literal2)
+
+    assert parse_version(parser, section, 'default') == version
+    assert parse_string(parser, section, 'string', 'default') == string
+    assert not parse_bool(parser, section, 'bool', 'True')
+    assert parse_literal(parser, section, 'literal', 'default') == ['a', 'b', 'c']
+    assert parse_literal(parser, section, 'literal2', 'default') == 1.23
+
+
+def test_environment_vars():
```
mircealungu, line 45: isn't this testing the `os` package?

tests/unit/database/test_endpoint.py
```diff
@@ -0,0 +1,41 @@
...
+@pytest.mark.parametrize('endpoint__monitor_level', [1])
+def test_update_endpoint(session, endpoint):
+    """Test whether the function returns the right values."""
+    update_endpoint(session, endpoint.name, 2)
+    assert get_endpoint_by_name(session, endpoint.name).monitor_level == 2
+
+
+@pytest.mark.parametrize('timestamp', [datetime(2020, 2, 2), datetime(2020, 3, 3)])
+def test_update_last_accessed(session, endpoint, timestamp):
+    """Test whether the function returns the right values."""
```
mircealungu, line 31: same unnecessary comment?

tests/unit/database/test_request.py
thejager: I agree, to me these comments fall in the same category as comments like:

```
# assign 9 to a
a = 9
```

:)

thejager:
> One comment that I forgot to make is that we don't test the interaction of our frontend with the backend (JS code that makes the requests and Python that handles it). The reason for leaving this out of the PR is that I'm not super familiar with this. Also, I don't really know what's the best approach here. Some plain JS tests, or full functional tests using Selenium?
> 
> @thejager What do you suggest?

I wouldn't bother with selenium test to be honest 😄If the API is tested, and the frontend is simple, not much can go wrong.

If anything, you could consider testing the controllers of the frontend. Mock the API and test whether the logic in the controller is correct.

flask_monitoringdashboard/main.py
FlyingBird95: Cool idea Johan, I'll create an examples folder in another pull request.

tests/fixtures/dashboard.py
FlyingBird95: I'm planning on updating user management in another PR, so I won't this now.

tests/unit/core/config/test_config.py
FlyingBird95: Removed

PR #338, opened 2020-05-23
Reports presenting difference between two commits, rather than two intervals

thejager:
A feature requested by Mircea.

image


flask_monitoringdashboard/database/request.py
```diff
@@ -8,33 +8,31 @@
...
-
+def get_latencies_sample(db_session, endpoint_id, criterion, sample_size=500):
     dialect = db_session.bind.dialect.name
 
     if dialect in ['sqlite', 'mysql']:
         order_by = func.random() if dialect == 'sqlite' else func.rand()
 
         items = db_session.query(Request.duration) \
-            .filter(Request.endpoint_id == endpoint_id, *criterion) \
+            .filter(Request.endpoint_id == endpoint_id, criterion) \
```
mircealungu, line 25: shouldn't you have *criterion here? 
or alternatively, above in get_latencies_in_timeframe not have a star?

thejager: Honestly, not sure yet. The asterisks are confusing me today 🙄

It seems to be working now though ...

bogdanp05:
When I try to compare a commit to itself, I don't see anything and I get the message in the console:
```
127.0.0.1 - - [02/Jun/2020 10:28:16] "POST /dashboard/api/reporting/make_report/commits HTTP/1.1" 200 -
No commit has been made, due to the following error: All values are below the grand median (201.51138305664062).
No commit has been made, due to the following error: 'NoneType' object has no attribute 'is_significant'
```

flask_monitoringdashboard/database/request.py
bogdanp05: The * unpacks iterable objects. There's an inconsistency now, because criterion in line 25 has to be a single criterion, while the same argument is passed to the function above, which expects a list of criterions (criteria?) in line 13.

flask_monitoringdashboard/controllers/requests.py
```diff
@@ -77,11 +79,15 @@ def get_status_code_frequencies(db_session, endpoint_id, *criterion):
...
-        .filter(Request.endpoint_id == endpoint_id, Request.status_code.isnot(None), *criterion) \
-        .group_by(Request.status_code).all()
+    status_code_counts = db_session.query(Request.status_code,
+                                          func.count(Request.status_code)) \
+        .filter(Request.endpoint_id == endpoint_id, Request.status_code.isnot(None),
+                criterion) \
+        .group_by(Request.status_code)
 
-    return dict(status_code_counts)
+    print(status_code_counts.all())
```
bogdanp05, line 88: this should be removed

thejager: Whoops :)

thejager:
> When I try to compare a commit to itself, I don't see anything and I get the message in the console:
> 
> ```
> 127.0.0.1 - - [02/Jun/2020 10:28:16] "POST /dashboard/api/reporting/make_report/commits HTTP/1.1" 200 -
> No commit has been made, due to the following error: All values are below the grand median (201.51138305664062).
> No commit has been made, due to the following error: 'NoneType' object has no attribute 'is_significant'
> ```

You shouldn't be able to compare a commit to itself, that makes no sense. I added some error checking!

mircealungu:
Johan, can you sort the commits by their datetime and display that also in the combobox? 

Otherwise ti's hard to use. 

Now: 

image


An example of how they'd look better: 

    https://api.zeeguu.org/dashboard/endpoint/8/versions

Probably even better with timestamp first, and then commit hash.

flask_monitoringdashboard/database/request.py
mircealungu: @thejager - any updates on this?

mircealungu:
@thejager - do you have time to look into the comment I made above? dank u wel!

thejager:
Hi @mircealungu I will have a look at this tomorrow!

mircealungu:
@thejager aaabsolutely!
if possible for the datetime to be first and then the hash it would be even more awesome :)

thejager:
done! :)

mircealungu:
cool! 
I wonder though: why are the CI tasks failing?

thejager:
@bogdanp05 @FlyingBird95 

```
No commit has been made, due to the following error: Neither 'BooleanClauseList' object nor 'Comparator' object has an attribute 'start_date'
```

Any tips on how I can debug this? I know an exception occurred with a db session, but there's not way of knowing where this exception occurred, right?

bogdanp05:
Hi @thejager ! There is actually. [Here](https://github.com/flask-dashboard/Flask-MonitoringDashboard/blob/master/flask_monitoringdashboard/database/__init__.py#L165) you can comment out the `except` block, so lines 165-167 to get the entire stack trace. Don't forget to uncomment it back after fixing the issue.

thejager:
> Hi @thejager ! There is actually. [Here](https://github.com/flask-dashboard/Flask-MonitoringDashboard/blob/master/flask_monitoringdashboard/database/__init__.py#L165) you can comment out the `except` block, so lines 165-167 to get the entire stack trace. Don't forget to uncomment it back after fixing the issue.

Thanks! That's much simpler than the solutions I was thinking of haha

PR #340, opened 2020-06-01
Feature/330 customize blueprint

### LINKED ISSUE #330: Blueprint name collision
**Describe the bug**
Hello,

I'm getting a blueprint name collision error.
```
AssertionError: A blueprint's name collision occurred between <flask.blueprints.Blueprint object at 0x0000021E18B91E48> and <flask.blueprints.Blueprint object at 0x0000021E16FA4940>.  Both share the same name "dashboard".  Blueprints that are created on the fly need unique names.
```
That's true because I have a route called "dashboard" already, but I changed the custom route link as described in the documentation as:

```
[dashboard]
APP_VERSION=1.0
CUSTOM_LINK=monitoring
MONITOR_LEVEL=3
OUTLIER_DETECTION_CONSTANT=2.5
SAMPLING_PERIOD=20
ENABLE_LOGGING=True

[authentication]
USERNAME=admin
PASSWORD=admin
GUEST_USERNAME=guest
GUEST_PASSWORD=['dashboardguest!', 'second_pw!']
SECURITY_TOKEN=cc83733cb0af8b884ff6577086b87909

[visualization]
TIMEZONE=Europe/Amsterdam
COLORS={'main':'[0,97,255]',
        'static':'[255,153,0]'}
```

my init looks as follows:

```
import flask_monitoringdashboard as monitoring_dashboard
app_service = Flask(__name__, )
monitoring_dashboard.config.init_from(file="../../monitoring_config.cfg")
monitoring_dashboard.bind(app_service)
app_service.config['JSON_SORT_KEYS'] = False
app_service.register_blueprint(routes.register.api, url_prefix='/register')
app_service.register_blueprint(routes.login.api, url_prefix='/login')
app_service.register_blueprint(routes.dashboard.api, url_prefix='/dashboard')
app_service.register_blueprint(routes.payment.api, url_prefix='/payment')
..
.
.
```

When I try to change the blueprint names myself like:

```
monitoring_dashboard.config.link = "monitoring"
monitoring_dashboard.blueprint.name = "monitoring"
```

it seems to be woring but when I call `localhost/monitoring/` I'm getting a:

`werkzeug.routing.BuildError: Could not build url for endpoint 'dashboard.index'. Did you mean 'dashboard.get_users_increase' instead?`

That's really weird, why does it not seem to work with the adjusted config file?

bogdanp05:
Support configurable FMD blueprint name via the env var FMD_BLUEPRINT. Fixes #330.
Btw, seeing how in the past 2 months there's been a lot of new stuff merged into develop, I decided to bump the version to 3.1.0 instead of 3.0.10.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,6 +3,4 @@ codecov:
 
 coverage:
   status:
-    patch:
-      default:
-        threshold: 1%
+    patch: off
--- a/config.cfg
+++ b/config.cfg
@@ -1,8 +1,7 @@
 [dashboard]
-APP_VERSION=3.3
+APP_VERSION=1.0
 GIT=/<path to your project>/.git/
-BLUEPRINT_NAME=foo
-CUSTOM_LINK=bar
+CUSTOM_LINK=dashboard
 MONITOR_LEVEL=1
 OUTLIER_DETECTION_CONSTANT=2.5
 
@@ -14,7 +13,7 @@ GUEST_PASSWORD=['dashboardguest!', 'second_pw!']
 SECURITY_TOKEN=cc83733cb0af8b884ff6577086b87909
 
 [database]
-;DATABASE=sqlite:////<path to your project>/dashboard.db
+DATABASE=sqlite:////<path to your project>/dashboard.db
 
 [visualization]
 TIMEZONE=Europe/Amsterdam
--- a/config.cfg
+++ b/config.cfg
@@ -1,8 +1,9 @@
 [dashboard]
-APP_VERSION=1.0
+APP_VERSION=3.3
 GIT=/<path to your project>/.git/
-CUSTOM_LINK=dashboard
-MONITOR_LEVEL=3
+BLUEPRINT_NAME=foo
+CUSTOM_LINK=bar
+MONITOR_LEVEL=1
 OUTLIER_DETECTION_CONSTANT=2.5
 
 [authentication]
@@ -13,7 +14,7 @@ GUEST_PASSWORD=['dashboardguest!', 'second_pw!']
 SECURITY_TOKEN=cc83733cb0af8b884ff6577086b87909
 
 [database]
-DATABASE=sqlite:////<path to your project>/dashboard.db
+;DATABASE=sqlite:////<path to your project>/dashboard.db
 
 [visualization]
 TIMEZONE=Europe/Amsterdam
--- a/config.cfg
+++ b/config.cfg
@@ -1,6 +1,7 @@
 [dashboard]
 APP_VERSION=1.0
 GIT=/<path to your project>/.git/
+BLUEPRINT_NAME=dashboard
 CUSTOM_LINK=dashboard
 MONITOR_LEVEL=1
 OUTLIER_DETECTION_CONSTANT=2.5
--- a/flask_monitoringdashboard/__init__.py
+++ b/flask_monitoringdashboard/__init__.py
@@ -26,7 +26,7 @@ def loc():
 
 
 config = Config()
-blueprint = None
+blueprint = Blueprint('dashboard', __name__, template_folder=loc() + 'templates')
 
 
 def bind(app, schedule=True):
@@ -36,8 +36,7 @@ def bind(app, schedule=True):
     :param app: the app for which the performance has to be tracked
     :param schedule: flag telling if the background scheduler should be started
     """
-    global blueprint
-    blueprint = Blueprint(config.blueprint_name, __name__, template_folder=loc() + 'templates')
+    blueprint.name = config.blueprint_name
     config.app = app
     # Provide a secret-key for using WTF-forms
     if not app.secret_key:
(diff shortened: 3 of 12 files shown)


config.cfg
```diff
@@ -2,7 +2,7 @@
 APP_VERSION=1.0
 GIT=/<path to your project>/.git/
 CUSTOM_LINK=dashboard
-MONITOR_LEVEL=3
+MONITOR_LEVEL=1
```
mircealungu, line 5: good!

codecov.yml
```diff
@@ -3,6 +3,4 @@ codecov:
 
 coverage:
   status:
-    patch:
-      default:
-        threshold: 1%
+    patch: off
```
mircealungu, line 6: why are we disabling this?
does this not disable codecov for all the PRs? 
is this what you want?

flask_monitoringdashboard/__init__.py
```diff
@@ -19,15 +19,14 @@
...
-
 
 def loc():
     """Get the current location of the project."""
     return os.path.abspath(os.path.dirname(__file__)) + '/'
 
 
-blueprint = Blueprint('dashboard', __name__, template_folder=loc() + 'templates')
+config = Config()
+blueprint = Blueprint(config.blueprint_name, __name__, template_folder=loc() + 'templates')
```
mircealungu, line 29: eheh. as the number of users grows, we'll end up extracting all the hardcoded strings into configuration options :)

flask_monitoringdashboard/core/config/__init__.py
```diff
@@ -27,6 +27,7 @@ def __init__(self):
         """Sets the default values for the project."""
         # dashboard
         self.version = '1.0'
+        self.blueprint_name = os.environ.get('FMD_BLUEPRINT', 'dashboard')
```
mircealungu, line 30: sorry, why don't we read this from the config file?

codecov.yml
bogdanp05: The message was `@codecov codecov/patch — 70.00% of diff hit (target 91.36%)` with a failed status. According to the [docs](https://docs.codecov.io/docs/commit-status#patch-status), the 70% comes from the coverage of the diff and it was required to be at least as high as the project coverage of 91.36%, which is already very high. I don't know if we want to enforce such a strict coverage requirement on all PRs.

FlyingBird95: I'm fine with removing the patch codecov 👍

flask_monitoringdashboard/core/config/__init__.py
FlyingBird95: I agree with mircea that this should come from the config file

bogdanp05: The `init_from` method of the `Config` class is called by the user after the FMD is imported. However, the blueprint is created in FMD's `__init__` itself, so we need the blueprint name there. If we want to read it from a file at import time we need to either: 
a) fix the location of the config file itself 
b) fix the name of the env var to point to the location of the config file.
Now I'm doing option b)  but minus the extra step of pointing to a file, I'm giving the value directly.

codecov.yml
mircealungu: Aha! 
Thanks for the clarification Bogdan!

flask_monitoringdashboard/core/config/__init__.py
mircealungu: ouch! i see now.

mircealungu: wouldn't it work if the blueprint would be initialized with None 
and then within the `.bind` that global blueprint would be overwritten with whatever's coming from config if there's something ?

bogdanp05: Good idea, it should work. I'll give it a try

bogdanp05: @mircealungu Turns out it's not a good idea to create a new blueprint object with every call of `bind`.  The first call of `bind` works fine, but then starting with the second call, the url's don't get registered anymore, hence the failing tests. Scratched my head for a while as to why, but in the end it was enough to just keep the global blueprint and only change its name in `bind`.

mircealungu: Changing the name sounds much more elegant! Nice :)

saravanakesavan:
Hey @bogdanp05, @FlyingBird95 

Is this changes available in 3.1.0 release of Flask-MonitoringDashboard?

saravanakesavan:
Hi @FlyingBird95 @mircealungu @bogdanp05 ,

I am trying to give a BLUEPRINT_NAME in config. While trying that the UI page is not loading. 

http://localhost:5001/xyzdashboard

I am getting the following error:

raise BuildError(endpoint, values, method, self)
werkzeug.routing.BuildError: Could not build url for endpoint 'xyzdashboard.index'. Did you mean 'dashboard.index' instead?

PR #347, opened 2020-06-14
User management

FlyingBird95:
Hey,

Please take your time to review another big pull request. This PR is about the user management of the FMD. Instead of defining hard-coded usernames and password in the configuration file as we're doing up to now, there is now an option to create users in the dashboard directly using a simple UI.

To recap, the FMD allows two roles: admins, who can do everything, and non-admins (aka guests), who can only see data.

I've made the following architectural decisions:
1. All users are displayed in the configuration page of the dashboard. Only admin users can see the list with all configured usernames in the database. Guests cannot see this.
2. An admin is allowed to delete any other user, so also other admins. The only user you cannot delete is yourself. Otherwise you would block your own access to the FMD.
3. An admin can create new users (either guest users or admin users).
4. An admin can update existing users. There are two types of updates:
  - Updating a user by changing the `is_admin` permission to `True`/`False`. I've restricted that the admin (who is logged into the FMD) cannot remove admin permissions from itself.
 - Updating a user by changing the password. The admin can change every user's password, as long as he knows the current password of that user.

Further comments:
- Passwords are hashed in the database, I've used two functions from `werkzeug.security`: `User.generate_password_hash(password)`, and `User.check_password_hash(password)`.
- The configuration-file still allows to set a username and password. This is the default in case the User-table is empty. The config has a default of `username=admin` and `password=admin`. 
So, in case someone starts with a fresh DB (without config), he can log into the FMD using these credentials and modify them in the configuration-page. In case someone already has an existing configuration, then he can log in with the username/password that is defined in the configuration-file. So, there is no data migration needed (for admin-accounts). Previously configured guest accounts are lost, because we don't support this configuration anymore.
- I've created an `angularjs`-directive, which is `modal`. This is pretty nice, because we can now do something like this:
```
<modal name="'logout'" title="'Logout?'">
    <p>Are you sure you want to logout?</p>
</modal>
```
Which renders like this.
Screenshot 2020-06-14 at 22 25 22

Note that I've used two types of quotes in the `name` and `title` argument, since angularjs interprets `name="logout"` as the variable `logout` (this took me a while to debug 😓 ).

Thanks in advance for reading all this. Now you can start on the code 😅

Kind regards,

Patrick

p.s. @thejager do you want to comment on the frontend code, and see if my changes make sense?


flask_monitoringdashboard/database/auth.py
```diff
@@ -0,0 +1,34 @@
...
+    In case the User-table is empty, a user with default credentials is added.
+    """
+    with session_scope() as session:
+        if session.query(User).count() == 0:
+            user = User(username=config.username, is_admin=True)
+            user.set_password(password=config.password)
+            session.add(user)
+
+        user = session.query(User).filter(User.username == username).one_or_none()
+        if user is not None:
```
bogdanp05, line 16: or `if user:`

flask_monitoringdashboard/static/elements/modal.html
```diff
@@ -0,0 +1,27 @@
+<div class="modal fade" id="{{ name }}Modal" tabindex="-1" role="dialog"
```
thejager, line 1: Personally, I don't think this is a generalisation worth the effort.

Soon enough you're going to need to want to add an icon to the Save button, customise the styling of the error message, etc... Then you're going to have to either take out this code and paste it back to where it's used or pass a ton of attributes.

Anyway, I'm not saying it's wrong, but I wouldn't have done it. I would have made a live template in my IDE and just generate the modals. Something like this:


Screen Recording 2020-06-22 at 15 13 55

flask_monitoringdashboard/static/pages/configuration.html
```diff
@@ -43,6 +35,97 @@
     </div>
 </div>
 
+<modal name="'delete'" title="'Delete user'">
```
thejager, line 38: For these kinds of confirmations, I would personally just use something like [sweetalert](https://sweetalert2.github.io/) or if I am feeling really lazy (90% of my projects), just roll with:

```js
window.confirm("Are you sure you want to delete this user?") && deleteUser();
```

flask_monitoringdashboard/frontend/js/controllers/configuration.js
```diff
@@ -1,14 +1,101 @@
-export function ConfigurationController($scope, $http, menuService, endpointService) {
+export function ConfigurationController($scope, $http, menuService, endpointService, modalService) {
     endpointService.reset();
     menuService.reset('configuration');
+    modalService.setConfirm('delete', () => deleteUser($scope.user));
```
thejager, line 4: Bit of an awkward API if you ask me. A result of (if you ask me) a bad generalisation of modals :) (see below)

flask_monitoringdashboard/static/pages/configuration.html
FlyingBird95: I would like to keep the modal, because it looks better 😅

PR #360, opened 2020-09-03
Development

FlyingBird95:
Time for releasing version 3.1.0 to live.
Short summary of the changes:
- Added support for Python 3.8
- Started using pytest instead of python's unittest
- Started using Webpack for frontend packaging
- Improved reports
- Moved to Github Actions from Travis for CI/CD
- Improved docstrings
- Various bug fixes


flask_monitoringdashboard/frontend/js/controllers/configuration.js
```diff
@@ -0,0 +1,98 @@
...
+            'api/user/create',
+            $.param({
+                'username': $('#create-username')[0].value,
+                'password': $('#create-pwd')[0].value,
+                'password2': $('#create-pwd2')[0].value,
+                'is_admin': $('#create-admin')[0].checked,
+            }),
+            {
+                headers: {
+                    'Content-Type': 'application/x-www-form-urlencoded;charset=utf-8;'
```
mircealungu, line 46: the most minor of the minor observations: shouldn't this be a constant somewhere?

mircealungu:
shouldn't the Readme.md be updated?
where now it says that to install one has to do `setup.py install` shouldn't there be also a webpack run before?

flask_monitoringdashboard/database/count_group.py
```diff
@@ -6,47 +6,47 @@
...
         .group_by(Request.endpoint_id)
         .all()
     )
 
 
-def get_value(list, name, default=0):
+def get_value(iterable, name, default=0):
     """
-    :param list: must be structured as: [(a, b), (c, d), ..]
+    :param iterable: must be structured as: [(a, b), (c, d), ..]
```
bogdanp05, line 27: I'd call this `tuples`, as it expects a list of tuples, not any iterable.

flask_monitoringdashboard/database/request.py
```diff
@@ -8,39 +8,28 @@
...
+    dialect = session.bind.dialect.name
 
-        items = db_session.query(Request.duration) \
-            .filter(Request.endpoint_id == endpoint_id, *criterion) \
-            .order_by(order_by) \
-            .limit(sample_size) \
-            .all()
+    if dialect == 'sqlite':
+        query = query.order_by(func.random())
+    elif dialect == 'mysql':
```
bogdanp05, line 21: what about other dialects? The SA answer states that it should be possible for Postgresql and Oracle DB

flask_monitoringdashboard/frontend/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
```
bogdanp05, line 1: I'd say move this to the main .gitignore. One git repo, one .gitignore.

flask_monitoringdashboard/database/request.py
FlyingBird95: This function produces a different query on different databases, and this is something that we should try to avoid as much as possible, as this will introduce unexpected errors. However, in this case, only the ordering is slightly different for different databases, and I don't see this as a problem.

FlyingBird95:
> shouldn't the Readme.md be updated?
> where now it says that to install one has to do `setup.py install` shouldn't there be also a webpack run before?

We don't need to run a webpack run before installing the fmd, as we ship the dist folder as well.
This should also fix issue #311

I've updated the code to address the remaining comments

bogdanp05 (approved):

mircealungu:
One more 👍  : I've deployed 3.1.0 on the Zeeguu API and 
things are fine :)

However, I thought the version comparison issue was 
fixed at some point? 

image

FlyingBird95:
Good spot @mircealungu . I have taken a look at the code that compares the Pypi version with the dashboard version, and found an error in this code. Try spotting the error yourself here: https://github.com/flask-dashboard/Flask-MonitoringDashboard/blob/0afd189b89648d41ea04bdfb1c6facc18f9617ff/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js#L68
And see my changes here: https://github.com/flask-dashboard/Flask-MonitoringDashboard/pull/360/commits/d4176938b17dfd0067178a5eb146f9caea892fc3#diff-0f86b4e611ed41735e8fc0dc04d2f9dd

FlyingBird95:
@mircealungu @bogdanp05 After we released version 3.1.0 to production, we should take a look at the issues that we have, as I think that a lot of them are already solved in this PR.

mircealungu:
@FlyingBird95 - good idea about having a new look at the issues :)

PR #371, opened 2021-02-03
Feature/tighter UI

Qblack:
Hi - 
I found this tool and I love it but I plan to use it for something about more microservicesy. 
I found it a bit too spacy when I ended up with  multiple endpoints worth of apps. 
So I made a couple changes which I hope other people will like. 
Ignore the blueprint field - I'm assuming that should be another PR. 

image

### DIFF
--- a/flask_monitoringdashboard/frontend/sass/custom.css
+++ b/flask_monitoringdashboard/frontend/sass/custom.css
@@ -508,3 +508,9 @@ footer.sticky-footer {
     vertical-align: super;
     font-size: small;
 }
+
+@media (min-width: 1500px) {
+    .container {
+        max-width: 1600px !important;
+    }
+}
--- a/flask_monitoringdashboard/frontend/sass/custom.css
+++ b/flask_monitoringdashboard/frontend/sass/custom.css
@@ -509,7 +509,7 @@ footer.sticky-footer {
     font-size: small;
 }
 
-@media (min-width: 1500px) {
+@media (min-width: 1644px) {
     .container {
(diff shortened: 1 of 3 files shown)


alexander34ro:
@mircealungu I checked the changes and they look good. I see no issues in merging this.

alexander34ro (approved):

flask_monitoringdashboard/frontend/sass/custom.css
```diff
@@ -507,4 +507,10 @@ footer.sticky-footer {
     color: red;
     vertical-align: super;
     font-size: small;
+}
+
+@media (min-width: 1500px) {
+    .container {
+        max-width: 1600px !important;
```
mircealungu, line 514: shouldn't both these be the same?
or at least the max-width smaller than min-width?

mircealungu:
Did you make the table rows less tall? 
Without a before/after view it's hard for me to imagine the changes :(

Qblack:
> Did you make the table rows less tall?
> Without a before/after view it's hard for me to imagine the changes :(

More compact yeah - I plan to end up with a lot end points in this. Heres a screen shot from the live demo for a before

image

flask_monitoringdashboard/frontend/sass/custom.css
Qblack: Honestly I have no idea. ```@media (min-width: 1200px){.container{max-width:1140px}}``` looks like you are right though. I think I understood it backwards.
I changed the min-width to 1644px and it still looks good for me.

Issue #276, opened 2019-11-27
hits today but last accessed three weeks ago?


mircealungu:
There seems to be a bug somewhere. 
Look at the discrepancy between the existence of hits today and last accessed being three weeks ago!
![image](https://user-images.githubusercontent.com/464519/69723522-0c376400-111a-11ea-8a99-b1f175f0f38d.png)

Qblack:
Hi @mircealungu I noticed this as well and was taking a look. It looks like the last_requested field is only updated on cache flushes. 
Is this UI by any chance seperate from that app? 
Like do you have multiple apps sharing the same db?

PR #373, opened 2021-02-03
Feature/micro service support

Qblack:
Basically I wanted the ability to not need scipy or the UI stuff so I could have a lightweight package to put on microservices. 

I added a setup_micro.py that only takes what I think we need - as well as flush_cache on teardown. 
The requirements are also truncated.


mircealungu:
Hi @Qblack ! 
Nice to hear that you find the FMD useful :)

And thanks for the PR!
We'll try to look at it and its sisters ASAP!

Qblack:
Sounds good - I imagine this one will have some back and forth - the other 2 are much more straightforward.

requirements-micro.txt
```diff
@@ -0,0 +1,17 @@
+click
+apscheduler
+flask>=0.9            # for monitoring the web-service
+sqlalchemy>=1.1.9     # for database support
+configparser          # for parsing the config-file
+psutil                # for logging extra CPU-info
+colorhash             # for hashing a string into a color
+numpy                 # for computing median and other stats
+pytz                  # for timezone info
+tzlocal               # for figuring out the local timezone
```
mircealungu, line 10: if you look at the requirements.txt you'll see that it has changed and is now frozen to 2.0.

requirements-micro.txt
```diff
@@ -0,0 +1,17 @@
...
+
+# pinning this to avoid flask pulling in a later one which breaks python3.6
+jinja2==2.11.1; python_version < '3.7'
+jinja2; python_version >= '3.7'
+# for statistical analysis of latency measurements
+# scipy<1.6.0; python_version < '3.7'
+# scipy; python_version >= '3.7'
```
mircealungu, line 17: remove these two commented lines maybe?

setup_micro.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+with open(loc + '/requirements-micro.txt') as f:
+    required = f.read().splitlines()
+
+with open('flask_monitoringdashboard/constants.json', 'r') as f:
+    constants = json.load(f)
+
+setuptools.setup(
+    name="Flask-MonitoringDashboard-Micro",
```
mircealungu, line 23: Is it necessary to add a new project name in PyPI?

mircealungu: Beyond having a new project name, there's a lot of duplication between files now.
setup_micro.py and requirements_micro have to always be changed with the corresponding files, and that is not cool. Can you think of a better solution?

flask_monitoringdashboard/__init__.py
```diff
@@ -76,6 +78,11 @@ def bind(app, schedule=True):
 
     atexit.register(flush_cache)
 
+    if micro:
```
mircealungu, line 81: why do we need a teardown only on micro? 
shouldn't this be in there always?

mircealungu (commented):
@Qblack - I finally had the chance to look a bit at the PR. I really like the idea of deploying for micro-service architectures. How do you exactly do this microservices setup? It would be cool if you could add a FMD-Micro.md maybe in which you describe this? I think it might be good for others too!

I've left a few comments on the code. Overall I like the idea of allowing for a setup_micro that does not deploy scipy (I guess the gain there is speed of deployment?). But I don't like the code duplication between setup*.py and requirements*.txt files. Do you think we can fix that somehow?

flask_monitoringdashboard/__init__.py
Qblack: So this is here to help with https://github.com/flask-dashboard/Flask-MonitoringDashboard/issues/276. 
The reason it was happening for me was because I am using one dashboard for multiple apps and the cache flush only gets triggered on the overview call of the api. This meant the last call time was always out of date unless I went to that apps dashboard. 
I have it under micro because that is when the problem is most prevelent and assumed someone would want to test it properly in a non micro context.

setup_micro.py
Qblack: RE: Adding new pypi project - Probaly? I have my own that I use and it needed a seperate wheel for sure. I have no idea if there is a way to do versions of your project. I think i saw someone do weird stuff with []s before though so could be a possible avenue to persue.

For the other point you are right. We could in theory import setup.py into setup micro and then modify it appropriatly in setup_micro - I assume we can pass an exploded dict to .setup. Alternatively we could have a common file that gets imported and modified by both. 

Requirements wise - probably the same deal - have a common reqs file and then a micro and standard might work?

I kinda took the path of least resistance for this since I wanted to use it myself right away and assumed you would have some feedback.

requirements-micro.txt
Qblack: Right. I think that happend after I started working on this. Added.

requirements-micro.txt
Qblack: Done

FlyingBird95:
Hey @Qblack,

Thanks for showing interest in the FMD. I appreciate your enthousiasm in the project. 

About your PR, I have couple of general comments for you:

- First about the `requirements-micro.txt`: We indeed need two requirements files, but it is not ideal to list the dependencies in both of the files. As a solution, we can do the following:
  requirements-micro.txt:
  <list all requirements that are needed for the micro FMD>

  requirements.txt
  first line (this is the trick): add the following line `-r requirements-micro.txt`
  <list all requirements that the micro does not need>
- about the setup file. We should not have this as a separate file. There should be only one. I think that we can distinguish the install command by checking `sys.args`. I.e.:
  - to install the FMD normally: `python setup.py`.
  - to install the micro FMD: `python setup.py --micro`.

Can you address these comments?

I'm looking forward to your future PR's for the FMD.

Patrick

Qblack:
@FlyingBird95 
Thanks Patrick - I definetly like youre idea more than mine and it seems to work on my end.

Q

setup.py
```diff
@@ -1,10 +1,17 @@
...
+import sys
 
 import setuptools
 
 loc = os.path.dirname(os.path.abspath(__file__))
 
+micro = False
+if '--micro' in sys.argv:
+    micro = True
+    sys.argv.remove('--micro')
```
FlyingBird95, line 12: you don't need to remove it from the args

setup.py
```diff
@@ -13,18 +20,43 @@ def get_description():
         return info + '\n\n' + changelog.read()
 
 
-with open(loc + '/requirements.txt') as f:
+with open(loc + '/requirements-micro.txt') as f:
     required = f.read().splitlines()
+if not micro:
+    with open(loc + '/requirements.txt') as f:
+        for line in f:
```
FlyingBird95, line 27: this can be a one-liner:
`required.extend([l for l in f.read().splitlines() if l != '-r requirements-micro.txt'])`

setup.py
```diff
@@ -13,18 +20,43 @@ def get_description():
...
+        'flask_monitoringdashboard.core.config',
+        'flask_monitoringdashboard.core.custom_graph',
+        'flask_monitoringdashboard.core.profiler',
+        'flask_monitoringdashboard.core.profiler.*',
+    ],
+        exclude=['flask_monitoringdashboard.controllers.*',
+                 'flask_monitoringdashboard.frontend.*',
+                 'flask_monitoringdashboard.views.*',
+                 'flask_monitoringdashboard.static.*',
+                 'tests', 'test.*']
```
FlyingBird95, line 49: we should also exclude the tests in non-micro mode

setup.py
Qblack: image

I could not get it to work without removing it. I think the extra arg was getting passed into setup. Althought to be fair I also probably make the wheels differently than you guys. So maybe thats why.

setup.py
Qblack: Done

flask_monitoringdashboard/__init__.py
```diff
@@ -29,12 +29,13 @@ def loc():
 blueprint = Blueprint('dashboard', __name__, template_folder=loc() + 'templates')
 
 
-def bind(app, schedule=True):
+def bind(app, schedule=True, micro=False):
```
FlyingBird95, line 32: maybe we should name it `without_dashboard=False`?

flask_monitoringdashboard/__init__.py
```diff
@@ -29,12 +29,13 @@ def loc():
...
     """Binding the app to this object should happen before importing the routing-
     methods below. Thus, the importing statement is part of this function.
 
     :param app: the app for which the performance has to be tracked
     :param schedule: flag telling if the background scheduler should be started
+    :param micro: flag telling if the views should be added - default False - True if you want only the db stuff
```
FlyingBird95, line 38: I think this docstring deserves an update:
- ` - default False - `: can be removed.
- `True if you only want the db stuff` doesn't sound really professional to me :)

setup.py
Qblack: For whatever reason find_packages has been giving me pain lately. I had thought you were excluding tests via specifying the test suite later - but maybe not? 

Because of the explicit includes up above these are technically not needed and likely just got copied from wherever the last time i used exclude was. Are you able to double check that it currently includes them on your end? Something in the way I am building could be funky.

flask_monitoringdashboard/__init__.py
Qblack: I can do that - I tried a couple different names but figured I should have it match with everything else being called micro.

flask_monitoringdashboard/__init__.py
Qblack: Fair. removed.

FlyingBird95:
Nice work,

It looks good, so you have my approval.

I'm still not entirely sure about the flag `without_dashboard=False`. Maybe we should do `include_dashboard=True`. But I'll leave this up to you.

Thank you for contributing to the FMD, and looking forward to seeing more PR's from you.

Patrick

FlyingBird95 (approved):

Qblack:
@FlyingBird95 ooh I like that alot better that way the if does not need a not. I only went with something that defaulted to false to try and be as unintrusive as possible.

mircealungu:
Now I see that this PR is still not merged. 
@Qblack @FlyingBird95: Is there anything else to be done here, or should we merge it in?

Qblack:
> Now I see that this PR is still not merged.
> @Qblack @FlyingBird95: Is there anything else to be done here, or should we merge it in?

Go for it :D

PR #387, opened 2021-05-06
Fixing requirements.txt

FlyingBird95:
I noticed that the requirements are failing in github actions. I decided to fix it by dropping the pinned version of `jinja2`, and let pip figure it out for himself.


mircealungu (approved):
good idea~!

Issue #348, opened 2020-06-15
How to log real ip when behind reverse proxy?


void4:
I'm using nginx to multiplex between multiple flask/uwsgi servers. However, flask_monitoringdashboard only logs the IP of the request, which is 127.0.0.1

The following is my nginx.conf:

```
server {
        listen 443 ssl;
        server_name randomeps.com www.randomeps.com;
        ssl_certificate /etc/letsencrypt/live/randomeps.com/cert.pem;
        ssl_certificate_key /etc/letsencrypt/live/randomeps.com/privkey.pem;

        location / {
                proxy_pass http://0.0.0.0:9000;
                proxy_set_header Host $host;
                proxy_set_header X-Forwarded-Proto $scheme;
                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;

                include /etc/nginx/mime.types;
        }
}


```

ameydeshpande30:
#390 I have created PR for this.

PR #390, opened 2021-05-20
Added config varibale for IP_HEADER.

ameydeshpande30:
- This is useful in case of reverse proxy servers .
- This allows user to modify IP_HEADER according to use case.
#348


thejager:
Hi, thanks for this PR :) I think it would be cool to have something that takes into account reverse proxies, but I do have a few remarks:

- This doesn't work for people with multiple chained reverse proxies. See https://en.wikipedia.org/wiki/X-Forwarded-For . `X-Forwarded-For` can contain a list of IP addresses.
- I'm a bit worried about security. Can an untrusted client spoof the `X-Forwarded-For` header if you set the header to `X-Forwarded-For`?
- It bothers me a bit that the user has to configure this.

I mostly write PHP so I'm sorry I only have a PHP example, but the `getClientIp()` method in Symfony has so many checks to make sure the right IP gets returned, taking into account trusted proxies, the full spec for `X-Forwarded-For` etc. I'm afraid this implementation might blow up in someone's face. 

https://github.com/symfony/http-foundation/blob/5.4/Request.php#L824

mircealungu:
It seems that this problem is solved by https://github.com/flask-dashboard/Flask-MonitoringDashboard/pull/394 ?

mircealungu:
closing this as there's no more activity for much too long time.

PR #394, opened 2021-07-10
Support for Custom `get_ip` Injection

us88:
This PR adds support for a custom IP address getter.

It is not correct to use `REMOTE_ADDR` in every case, though it is a reasonable fallback. We achieve this by introducing a function called `inject_dependencies` inside the `Config` class. This is going to be useful for future PRs because it opens the door for a few other things, such as:

* Overriding `url_for` usage in favor of some user-defined logic (In our case, we have an application using a multi-host configuration, so we don't define `SERVER_NAME` in our Flask configuration, and `url_for` does not do the right thing)
* Decoupling the authentication layer completely from FMD in favor of user-defined functions to determine if:
  * There is a logged in user
  * There is a logged in admin

Awaiting your feedback!

### DIFF
--- a/flask_monitoringdashboard/core/config/__init__.py
+++ b/flask_monitoringdashboard/core/config/__init__.py
@@ -60,7 +60,7 @@ class Config(object):
         # dependencies
         self.get_ip = None
 
-    def inject_dependencies(self, get_ip=None, url_for=None):
+    def inject_dependencies(self, get_ip=None):
         """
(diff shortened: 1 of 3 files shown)


FlyingBird95:
Hey us88,

Thanks for you PR. I agree with the changes, and therefore you have my approval. Note that I changed the target branch from `master` to `development`, so that it is in line with our policy.

Regards,

Patrick

FlyingBird95 (approved):

mircealungu (approved):

PR #397, opened 2021-07-12
Bugfix: Fixed incorrect `dashboard/api` usage in `reporting.js`

us88:
The endpoint strings in reporting.js were still assuming that the API is hosted at `/dashboard/`.

I changed them so that they match the endpoint strings in other files (just `api/*`).

Changes also applied to the static app.js file. The diff is terrible though... Not sure if I was supposed to do this, I guess it's part of a pipeline you have in place?


mircealungu:
@thejager - this looks good to me! what do you think?

PR #408, opened 2021-12-31
FIX KEYERROR EXCEPTION

Amedeo91:
Hello,
for my monitoring dashboard, I want to define endpoints that are not Flask routes. 
Those are the only point where I have errors due to key errors exceptions.

Thanks,
Amedeo


flask_monitoringdashboard/controllers/endpoints.py
```diff
@@ -144,10 +144,11 @@ def set_endpoint_rule(session, endpoint_name, monitor_level):
...
-
-    add_decorator(get_endpoint_by_name(session, endpoint_name))
+    if config.app.view_functions.get(endpoint_name):
+        # Remove wrapper
+        original = getattr(config.app.view_functions[endpoint_name], 'original', None)
+        if original:
+            config.app.view_functions[endpoint_name] = original
+        session.commit()
+
+        add_decorator(get_endpoint_by_name(session, endpoint_name))
```
FlyingBird95, line 154: I think this should be un-indented. We always want to add a decorator

flask_monitoringdashboard/controllers/endpoints.py
```diff
@@ -144,10 +144,11 @@ def set_endpoint_rule(session, endpoint_name, monitor_level):
...
-        config.app.view_functions[endpoint_name] = original
-    session.commit()
-
-    add_decorator(get_endpoint_by_name(session, endpoint_name))
+    if config.app.view_functions.get(endpoint_name):
+        # Remove wrapper
+        original = getattr(config.app.view_functions[endpoint_name], 'original', None)
+        if original:
+            config.app.view_functions[endpoint_name] = original
+        session.commit()
```
FlyingBird95, line 152: this line was already in the old code, but I think that's incorrect, since we're not making database changes, so committing has no effect. Can you remove this line?

FlyingBird95:
Hey Amedeo91,

thanks for your contributions to the Flask-MonitoringDashboard. I've left two comments. Please adjust them. I also changed your target branch to our development branch.

flask_monitoringdashboard/controllers/endpoints.py
Amedeo91: Yes I will!

flask_monitoringdashboard/controllers/endpoints.py
Amedeo91: Actually not quite. For my case, I want to log endpoints that are not Flask routes. Hence, I do not need to make remove and then add the decorator. Does it make sense?

Amedeo91:
Hey FlyingBird95,

I reply to your comment and update the PR.
Thank you for your time!

FlyingBird95 (approved):

Amedeo91:
The changes in this PR are available also here: https://github.com/flask-dashboard/Flask-MonitoringDashboard/pull/409
Hence, we should not merge this if we merge the one above

FlyingBird95:
Let's merge #409 after it's ready

PR #423, opened 2023-02-07
Use Blueprint.record_once instead of Blueprint.before_app_first_request

FlorianRhiem:
The Blueprint decorator method `before_app_first_request` is [deprecated](https://flask.palletsprojects.com/en/2.2.x/api/#flask.Blueprint.before_app_first_request) and will be removed in Flask 2.3. Instead, setup code should use the `record_once` decorator method. The functions wrapped with this decorator will be called when the blueprint is registered on an application instead of before the first request.


mircealungu (approved):
Thanks @FlorianRhiem !

PR #434, opened 2023-08-22
Fix compatibility with Flask 2.3.x

Voltini:
I just changed two lines in the __init__ file by following the solution on [this stack overflow post](https://stackoverflow.com/a/74629704).


mircealungu:
Thanks @Voltini ! This looks good. I'll test it a bit later, and if it works I'll merge it in and create a new release as soon as possible!

Voltini:
Hi @mircealungu I did some digging in the repo and found that someone else already submitted a different ~~and probably better~~ solution in #423, which has already been merged into master.

mircealungu:
You're right @Voltini !
I guess all we need is to release a new version corresponding to that change then!
I'll close this PR for now, thanks again.

PR #441, opened 2023-11-30
Updated read the docs file

JohannesLC:
Tested here to work:
https://fmd-weather.readthedocs.io/en/latest/functionality.html

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,12 +1,35 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
 version: 2
 
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
+    # You can also specify other tool versions:
+    # nodejs: "20"
+    # rust: "1.70"
+    # golang: "1.20"
+
+# Build documentation in the "docs/" directory with Sphinx
 sphinx:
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Please note that the changes before version 1.10.0 have not been documented.
 
-v3.2.1
```
wielas, line 8: Why don't we separate the changes? What actually changed between 3.2.1(0) and 3.2.2?

.readthedocs.yaml
```diff
@@ -1,12 +1,35 @@
...
 
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
+    # You can also specify other tool versions:
+    # nodejs: "20"
+    # rust: "1.70"
+    # golang: "1.20"
```
wielas, line 15: Kind of a noise. These are documentation notes, we don't need them here.

wielas (changes requested):

wielas (approved):

docs/changelog.rst
JohannesLC: Agreed, updated

.readthedocs.yaml
JohannesLC: It's the standard from ReadTheDocs. Suggest keeping it in

PR #442, opened 2023-12-02
debug and better design for mobile phones

amirhnir:
1. the **dashboard link** on the configuration file was broken so I fixed this
2. I made some changes for better design in devices with small widths for example the nav bar didn't show on mobile phones so I tried to fix this.


wielas (approved):

wielas:
Nice fix, thanks for taking effort!

Issue #437, opened 2023-11-15
Initial setup fails 


Cornul11:
Firstly, thanks to the maintainers for their great work!

**Describe the bug**
Installing and setting up the dashboard leads to an Exception after logging in.

Moreover, on the Configuration page, the "First request in the currently deployed version" field shows; aN-aN-NaN aN:aN, which is probably related.

**To Reproduce**
Steps to reproduce the behaviour:
1. Install the dashboard
2. Bind it to the app
3. Open the dashboard
4. Log in
5. See the main overview graph endlessly loading

**Expected behaviour**
It is quite clear that the expected behaviour is to have the dashboard overview page loading properly.

**Screenshots**
NA

**Desktop (please complete the following information):**
 - OS: Microsoft Windows 11
 - Browser chrome
 - FMD Version: 2.1.3

**Additional context**
Here is the console log when the issue arises:
```
127.0.0.1 - - [15/Nov/2023 09:40:54] "GET /dashboard/api/deploy_config HTTP/1.1" 200 -
127.0.0.1 - - [15/Nov/2023 09:41:12] "GET /dashboard/api/overview HTTP/1.1" 200 -
[2023-11-15 09:41:13,014] ERROR in app: Exception on /dashboard/api/deploy_details [GET]
Traceback (most recent call last):
  File "\flaskProject\venv\lib\site-packages\flask\app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
  File "\flaskProject\venv\lib\site-packages\flask\app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "\flaskProject\venv\lib\site-packages\flask_restful\__init__.py", line 298, in error_router
    return original_handler(e)
  File "\flaskProject\venv\lib\site-packages\flask_cors\extension.py", line 176, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "\flaskProject\venv\lib\site-packages\flask\app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "\flaskProject\venv\lib\site-packages\flask\app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "\flaskProject\venv\lib\site-packages\flask_monitoringdashboard\core\auth.py", line 39, in wrapper
    return func(*args, **kwargs)
[...]

Issue #435, opened 2023-09-21
Flask Dashboard not working, just screen loading and loading


leetanshaj:
**Screenshots**
<img width="1574" alt="Screenshot 2023-09-21 at 6 49 39 PM" src="https://github.com/flask-dashboard/Flask-MonitoringDashboard/assets/52253745/689b81cb-94cc-468f-925c-fc4a92e99a81">

What do I do here, facing this on my macOs, ubuntu20.04 with python3.8 python3.10 python3.9 and flask==2.2.3

mircealungu:
Do you see any errors in the terminal or the logs? It seems like the dashboard is not finding your endpoints. Is your codebase open-source. Alternatively can you isolate the minimum code that replicates the issue?

leetanshaj:
1. I've tried with a very basic app, it's not working
app=Flask(__name__)
@app.route('/test')
def kk():
    return {"ok":1}
    
2. No error in terminal, I've even turned on logging by dashboard.config.enable_logging=True, no error there
3. I checked the .db file, it contains table but table does not contain any row

mircealungu:
I'm missing the FMD `bind` call. Where do you do that?

leetanshaj:
`import flask_monitoringdashboard as dashboard
from flask import Flask

app = Flask(__name__)
dashboard.config.enable_logging=True
dashboard.bind(app)


@app.route("/test", methods = ['GET'])
def kk():
	return {"ok": True}


app.run()`

mircealungu:
ehi @leetanshaj .

altthough i didn't remember this being necessary, it seems that you have to run `dashboard.bind` after all the app routes have been defined. can you check that this code works for you?

```
import flask_monitoringdashboard as dashboard
from flask import Flask

app = Flask("name")


@app.route("/test", methods = ['GET'])
def kk():
	return {"ok": True}

dashboard.config.enable_logging=True
dashboard.bind(app)

app.run()
```

leetanshaj:
Thanks man! It works

mat-l:
Can confirm that "dashboard.bind(app)" must be executed after the routes.

PR #444, opened 2023-12-16
The example in the Setup section was edited to avoid errors.

amirhnir:
According to #437 and  #435 , I edited the example in the Setup section to avoid the error mentioned here: #437


mircealungu (approved):
thanks @amirhnir ! I don't know why I didn't integrate this sooner!

PR #448, opened 2024-03-11
Fixed create edit delete and checkboxes

JohannesLC:
Create user, delete user and edit user modals were broken, so I fixed them.

Additionally the checkboxes were showing up as field inputs, so I removed class formcontrol.


mircealungu (approved):
tak @JohannesLC !

Issue #449, opened 2024-03-14
User table - insufficient column length for password_hash


bhodges-navapbc:
**Describe the bug**
User creation fails when using a version of the [Werkzeug dependency >= v3.0.0](https://werkzeug.palletsprojects.com/en/3.0.x/changes/#version-3-0-0)

Flask-MonitoringDashboard uses generate_password_hash, which has switched to defaulting to scrypt in werkzeug.security. This results in hashes with a length of 163, whereas the User table password_hash column has a length of 128.

**To Reproduce**

Perform a base install and attempt to login with the default credentials. An account is not created in the User table.

Example error output using PostgreSQL for the database would include "value too long for type character varying(128)"

mircealungu:
Thanks for detecting the problem @bhodges-navapbc ! So the solution seems to be simply redefining the User table. Can you do that and send a PR? We'll integrate it right away if you could!

PR #450, opened 2024-03-14
Increase User.password_hash column size

bhodges-navapbc:
Increases the User.password_hash column size to 162. This accommodates a recent werkzeug.security update to generate_password_hash, which moved to scrypt and defaults to 16 for the salt length.

Additional details may be found in #449.


mircealungu (approved):
thanks @bhodges-navapbc !
we'll publish a new release soon!

Issue #451, opened 2024-03-14
Type not specified as int for endpoint_id get parameter in routes


bhodges-navapbc:
**Describe the bug**
Due to a type mismatch within a generated sql query, graphs aren't populating at the endpoint-level, when postgres is used as the database. sqlite seems to be more forgiving and I couldn't reproduce there.

An example exception:

```
No commit has been made, due to the following error: (psycopg.errors.UndefinedFunction) operator does not exist: integer = character varying
LINE 3: WHERE "Request".endpoint_id = $1::VARCHAR AND "Request".time...
                                    ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
[SQL: SELECT "Request".time_requested AS "Request_time_requested" 
FROM "Request" 
WHERE "Request".endpoint_id = %(endpoint_id_1)s::VARCHAR AND "Request".time_requested >= %(time_requested_1)s::TIMESTAMP WITHOUT TIME ZONE AND "Request".time_requested <= %(time_requested_2)s::TIMESTAMP WITHOUT TIME ZONE]
[parameters: {'endpoint_id_1': '11', 'time_requested_1': datetime.datetime(2024, 2, 29, 0, 0), 'time_requested_2': datetime.datetime(2024, 3, 14, 23, 59, 59)}]
(Background on this error at: https://sqlalche.me/e/20/f405)
```

**To Reproduce**

- Perform a base install, using postgres as the database.

- Login, then click on an endpoint. Graphs won't populate, for example hitting /dashboard/endpoint/11/hourly_load

**Expected behavior**
Graphs would populate at the endpoint-level.

**Additional context**
I'll submit a PR for this shortly.

PR #452, opened 2024-03-14
Specifies int type for endpoint_id get params in routes

bhodges-navapbc:
These changes ensure endpoint_id is of type int when passed as a GET parameter. For full details, please see issue #451.


mircealungu (approved):
This looks cool. Thanks for the fix @bhodges-navapbc !

PR #458, opened 2024-04-27
Update python-publish.yml

mircealungu:
the `deploy:needs:[test]` should not publish when the tests are failing cf. https://stackoverflow.com/a/63148947/1200070

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -6,9 +6,36 @@ on:
       - master
 
 jobs:
+
+  test:
+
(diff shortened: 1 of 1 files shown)


mircealungu:
@JohannesLC - can you double check this? it looks good to me, but many eyes catch all the bugs :)

JohannesLC:
Hey @mircealungu 
Looks good to me, but I tried running it locally and I seem to get unittests that fail and flask_monitoringdashboard/core/logger.py seem to give a circular reference, so we might need to fix those.

PR #468, opened 2025-02-04
Feature/sort endpoint table

Carmish:
Implement sorting of endpoints in Dashboard Overview table (both for Number of hits and Median request duration (ms)). 

Before:
Skærmbillede 2025-02-04 kl  17 38 55

After:
Skærmbillede 2025-02-04 kl  17 40 02

### DIFF
--- a/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
+++ b/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
@@ -5,8 +5,9 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
     $scope.pypi_version = '';
     $scope.dashboard_version = '';
     $scope.isHits = true;
+
     $scope.sortBy = 'name';
-    $scope.sortingOrder = {};
+    $scope.isDesc = true;
 
     $scope.table = [];
     $scope.selectedItem = 2;
@@ -30,12 +31,16 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
         $scope.table = response.data;
     });
 
-    function sortOrder(a, b){
+    function ascendingOrder(a, b){
       return a[$scope.sortBy] > b[$scope.sortBy] || b[$scope.sortBy] === null;
+    }
+
+    function descendingOrder(a, b){
+        return ascendingOrder(b, a);
     }
     
     function sortItems(items){
-      return $scope.getOrAddSortingOrder($scope.sortBy) ? items.sort(sortOrder) : items.sort((a, b) => sortOrder(b, a))
+      return $scope.isDesc ? items.sort(descendingOrder) : items.sort(ascendingOrder)
     }
 
     function getItemsForPage(pageNumber) {
@@ -52,19 +57,13 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
         return sortItems(items).slice(start, end);
     }
 
-    $scope.getOrAddSortingOrder = function (column) {
-        let orderBy = $scope.sortingOrder[column];
-        if (orderBy !== undefined) {
-            return orderBy;
-        }
-        $scope.sortingOrder[column] = true;
-        return true;
-    }
-    
     $scope.changeSortingOrder = function (column) {
-        for (const [key, value] of Object.entries($scope.sortingOrder))
-            $scope.sortingOrder[key] = key === column ? !value : true;
-        $scope.sortBy = column;
+        if (column !== $scope.sortBy){
+            $scope.isDesc = true;
+            $scope.sortBy = column;
+            return;
+        }
+        $scope.isDesc = !$scope.isDesc;
     }
 
     $scope.getFilteredItems = function () {
@@ -72,12 +71,10 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
     }
 
     $scope.getSortArrowClassName = function (column) {
-      let isCurrentProp = $scope.sortBy === column;
-      let desc = $scope.getOrAddSortingOrder(column);
       return {
-        'rotate-up': desc,
-        'rotate-down': !desc,
-        'text-gray': !isCurrentProp 
+        'rotate-up': !$scope.isDesc && $scope.sortBy === column,
+        'rotate-down': $scope.isDesc && $scope.sortBy === column,
+        'text-gray': $scope.sortBy !== column 
       }
     }
 
--- a/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
+++ b/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
@@ -62,10 +62,9 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
     }
     
     $scope.changeSortingOrder = function (column) {
-        let orderBy = $scope.getOrAddSortingOrder(column);
-        $scope.sortingOrder[column] = !orderBy;
+        for (const [key, value] of Object.entries($scope.sortingOrder))
+            $scope.sortingOrder[key] = key === column ? !value : true;
         $scope.sortBy = column;
-        return orderBy;
     }
 
     $scope.getFilteredItems = function () {
--- a/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
+++ b/flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
@@ -5,7 +5,8 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
     $scope.pypi_version = '';
     $scope.dashboard_version = '';
     $scope.isHits = true;
-    $scope.sortBy = { 'prop': 'name', 'desc': true}
+    $scope.sortBy = 'name';
+    $scope.sortingOrder = {};
 
     $scope.table = [];
     $scope.selectedItem = 2;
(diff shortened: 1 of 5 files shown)


mircealungu (changes requested):
thanks for adding the sorting to the table @Carmish! a few minor observations and one small refactoring request and then we're ready to merge!

flask_monitoringdashboard/frontend/webpack.config.js
```diff
@@ -16,26 +20,27 @@ module.exports = {
...
                     },
-                    'sass-loader'
-                ]
+                    "sass-loader",
+                ],
             },
             {
                 test: /\.(woff(2)?|ttf|eot|svg|otf)(\?v=\d+\.\d+\.\d+)?$/,
-                type: 'asset/resource',
+                type: "asset/resource",
```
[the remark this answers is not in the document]
AlbertRossJoh: the webpack configuration has been changed back, we were figuring out how to run the application

flask_monitoringdashboard/frontend/webpack.config.js
```diff
@@ -1,13 +1,17 @@
...
-    entry: ['./js/app.js', './sass/app.scss'],
+    mode: process.env.NODE_ENV === "production" ? "production" : "development",
+    entry: ["./js/app.js", "./sass/app.scss"],
     output: {
-        path: path.resolve(__dirname, '../static'),
-        filename: 'js/app.js',
-        publicPath: '/static/'
+        path: path.resolve(__dirname, "../static"),
+        filename: "js/app.js",
+        //publicPath: "/static/",
```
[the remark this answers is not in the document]
AlbertRossJoh: removed :D

flask_monitoringdashboard/frontend/webpack.config.js
```diff
@@ -16,26 +20,27 @@ module.exports = {
...
     plugins: [
         new MiniCssExtractPlugin({
-            filename: 'css/[name].css',
-        })
-    ]
+            filename: "css/[name].css",
+        }),
+    ],
 };
+console.log(module.exports);
```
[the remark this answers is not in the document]
AlbertRossJoh: removed

flask_monitoringdashboard/static/js/app.js
```diff
@@ -16,7 +16,7 @@
 /***/ ((__unused_webpack_module, __webpack_exports__, __webpack_require__) => {
```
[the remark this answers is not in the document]
AlbertRossJoh: the autogenerated code has now been removed from the repo, this should probably be in a separate PR, as this PR is delimited to the sorting of tables, what do you think?

flask_monitoringdashboard/static/pages/overview.html
```diff
@@ -18,71 +18,141 @@
...
+                      <table class="table table-hover table-bordered table-sm">
+                          <thead>
+                          <tr>
+                              <th></th>
+                            <th 
+                              ng-click="sortBy = {'prop': 'name', 'desc': !sortBy.desc}" 
+                              style="cursor: pointer">
+                                <div class="d-flex justify-content-between">
+                                  Endpoint
+                                  <span ng-class="{'rotate-up': sortBy.desc && sortBy.prop === 'name', 'rotate-down': !sortBy.desc && sortBy.prop === 'name', 'text-gray': sortBy.prop !== 'name'}">&uarr;</span>
```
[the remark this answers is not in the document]
AlbertRossJoh: the code is now much cleaner!

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
```diff
@@ -28,23 +30,57 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
         $scope.table = response.data;
     });
 
-    function getItemsForPage(pageNumber) {
-        const start = Number(pageNumber) * $scope.pageSize;
-        const end = (Number(pageNumber) + 1) * Number($scope.pageSize);
+    function sortOrder(a, b){
+      return a[$scope.sortBy] > b[$scope.sortBy] || b[$scope.sortBy] === null;
```
mircealungu, line 34: should this be called `descendingOrder`? Also, why is the second test only on b[sortBy] being null?

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
```diff
@@ -28,23 +30,57 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
...
+    }
+    
+    function sortItems(items){
+      return $scope.getOrAddSortingOrder($scope.sortBy) ? items.sort(sortOrder) : items.sort((a, b) => sortOrder(b, a))
```
mircealungu, line 38: Hard to understand this. Why do we call `sortOrder(b,a)`? Is this related to my above suggestion of naming the function `descendingOrder`? This is why we call it with `(b,a)`? And this is why we do the test no b being null first?

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
```diff
@@ -28,23 +30,57 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
...
+        return sortItems(items).slice(start, end);
+    }
 
-        return items.slice(start, end);
+    $scope.getOrAddSortingOrder = function (column) {
+        let orderBy = $scope.sortingOrder[column];
+        if (orderBy !== undefined) {
+            return orderBy;
+        }
+        $scope.sortingOrder[column] = true;
```
mircealungu, line 60: it's hard to read this. could it be that the name of the `sortingOrder` variable should be something else? e.g. `isSortedBy`? after all it seems we're setting it as a boolean, so it would be natural to hame it as a predicate. Or is it tracking whether it's ascending or descending? In that case it would be easier to follow if two constants would be introduced to track that, e.g. `sortingOrder = ASCENDING` vs `DESCENDING`. Or simply name the variable `isSortedAscending`?

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
```diff
@@ -28,23 +30,57 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
...
         if ($scope.slectedBlueprint) {
-            items = items.filter(item => item.blueprint===$scope.slectedBlueprint)
+            items = items.filter(item => item.blueprint===$scope.slectedBlueprint);
         }
+    
+        return sortItems(items).slice(start, end);
+    }
 
-        return items.slice(start, end);
+    $scope.getOrAddSortingOrder = function (column) {
```
mircealungu, line 55: `getOrAdd` is very hard to parsre. Can we simply call it `getSortingOrder`? and then as a comment add the observation that in case there info is not there, it's initalized?

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
```diff
@@ -28,23 +30,57 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
...
+            $scope.sortingOrder[key] = key === column ? !value : true;
+        $scope.sortBy = column;
     }
 
     $scope.getFilteredItems = function () {
         return getItemsForPage($scope.currentPage);
     }
 
+    $scope.getSortArrowClassName = function (column) {
+      let isCurrentProp = $scope.sortBy === column;
```
mircealungu, line 75: could this be named`isCurrentSortingCriterion`? I'm not sure what `prop` means in this context.

mircealungu (changes requested):
nice! a few more clarification questions, and suggestions of possible refactorings in there.

mircealungu:
one more thing to discuss when we meet:  should `main.css`, and `app.css` be versioned? it feels wrong to commit generated code.

AlbertRossJoh:
> one more thing to discuss when we meet: should `main.css`, and `app.css` be versioned? it feels wrong to commit generated code.

I have removed the main.css and app.css, I agree that whatever webpack generates should not be committed, as that could lead to conflicts further down the road.

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
AlbertRossJoh: I am not sure I agree as it adds the column if it does not exist in the object.

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
AlbertRossJoh: We do a null check as we define null as being less than everything else, I will fix the naming.

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
AlbertRossJoh: Right I can unpack this.
> Why do we call sortOrder(b,a)? This is why we call it with (b,a)? 

sortOrder(b,a), reverses the order, such that an array sorted as `[1, 2, 3]` becomes `[3, 2, 1]`

> Is this related to my above suggestion of naming the function descendingOrder?

I do not see how it relates to the `descendingOrder` refactor

>And this is why we do the test no b being null first?

I answered this in my reply above `We do a null check as we define null as being less than everything else, I will fix the naming.`

flask_monitoringdashboard/static/js/app.js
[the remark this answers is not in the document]
mircealungu: separate PR is fine!

AlbertRossJoh:
@mircealungu please review the updated code

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
```diff
@@ -28,23 +31,53 @@ export function OverviewController($scope, $http, $location, menuService, endpoi
         $scope.table = response.data;
     });
 
-    function getItemsForPage(pageNumber) {
-        const start = Number(pageNumber) * $scope.pageSize;
-        const end = (Number(pageNumber) + 1) * Number($scope.pageSize);
+    function ascendingOrder(a, b){
```
mircealungu, line 34: sorry folks, but this still looks like descending order to me. 
can you double check pls?

mircealungu (changes requested):

flask_monitoringdashboard/frontend/js/controllers/OverviewController.js
Carmish: You are right, the ascendingOrder was descending. It should be fixed now.

Regarding the null-test, If an endpoint has never been requested, and therefore has no Last requested date, we thought it made sense that null was ordered as smaller than any other timestamp.

mircealungu:
Yes! Then please add a comment!

PR #476, opened 2025-02-17
Detailed exception info is now available

AlbertRossJoh:
It is now possible to get detailed exception information on a specific endpoint, this means the usual exception info, but with an added stack trace such that it is visible where the exception occurred.


flask_monitoringdashboard/views/exception.py
NatalieClaraPetersen, line 1: <int:endpoint_id> instead like in the other endpoint api

NatalieClaraPetersen (changes requested):

PR #478, opened 2025-02-18
Removed generated files

AlbertRossJoh:
The autogenerated files from webpack has been removed. I believe that having these files can cause headaches as they can accidentally be committed, which will result in the great commit wars.


mircealungu (approved):
Tak!

PR #483, opened 2025-02-25
new db tables: ExceptionType and ExceptionMessage

NatalieClaraPetersen:


flask_monitoringdashboard/database/__init__.py
```diff
@@ -239,12 +239,30 @@ class CustomGraphData(Base):
...
     
     stack_lines = relationship('ExceptionStackLine', backref='exception_info')
+
+class ExceptionType(Base):
+    """Table for storing Exception types"""
+    
+    __tablename__ = '{}ExceptionType'.format(config.table_prefix)
+    
+    id = Column(Integer, primary_key=True)
+    type = Column(String(1500), nullable=False)
```
mircealungu, line 257: are we too generous with the length of this column? I think we are safe to keep it to 256 characters. if somebody uses exceptions longer than that, they can use a different exception monitor I'd say!

mircealungu: LeChat agrees: 
image

flask_monitoringdashboard/database/exception_info.py
```diff
@@ -31,12 +31,12 @@ def count_grouped_exceptions(session):
     """
     return (
         session.query(ExceptionInfo.request_id)
-        .join(Request, ExceptionInfo.request_id == Request.id)
-        .join(Endpoint, Request.endpoint_id == Endpoint.id)
+        .join(Request, ExceptionInfo.request)                  # Join Request via relationship
```
mircealungu, line 34: You're already ahead of me in joining with SqlAlchemy :)

mircealungu (approved):
I had a very cursory look and this looks good.

PR #484, opened 2025-02-25
You can now delete stack traces

AlbertRossJoh:
Screenshot 2025-02-25 at 18 39 28

### DIFF
--- a/flask_monitoringdashboard/database/exception_info.py
+++ b/flask_monitoringdashboard/database/exception_info.py
@@ -1,8 +1,10 @@
 """
 Contains all functions that access an ExceptionInfo object.
 """
+from os import walk
 from sqlalchemy import func, desc
-from flask_monitoringdashboard.database import ExceptionInfo, Request, Endpoint, ExceptionMessage, ExceptionType
+from flask_monitoringdashboard.database import ExceptionInfo, ExceptionStackLine, FullStackTrace, Request, Endpoint
+from flask_monitoringdashboard.core.database_pruning import delete_unreferenced_entries
 def get_exception_info(session, request_id: int):
     """
     Retrieve an ExceptionInfo record by request_id.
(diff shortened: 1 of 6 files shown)


flask_monitoringdashboard/database/exception_info.py
```diff
@@ -88,6 +90,20 @@ def get_exceptions_with_timestamps(session, offset, per_page):
...
+    :param full_stack_trace_id: the stack trace id
+    :return: None
+    """
+    session.query(ExceptionStackLine).filter(ExceptionStackLine.full_stack_trace_id == full_stack_trace_id).delete()
+    result = (session.query(FullStackTrace).filter(FullStackTrace.id == full_stack_trace_id).all())
+    for stack_trace in result:
+        session.delete(stack_trace)
+    delete_unreferenced_entries(session)
+    session.query(ExceptionInfo).filter(ExceptionInfo.full_stack_trace_id == full_stack_trace_id).delete()
+    session.commit()
```
NatalieClaraPetersen, line 106: delete_unreferenced_entries() does most of this (deleting ExceptionStackLine, FullStackTrace) and also FunctionDefinition which was overseen here, you just have to: 
delete the specific ExceptionInfo, then call delete_unreferenced_entries()

NatalieClaraPetersen (changes requested):

flask_monitoringdashboard/static/pages/exception_stack_trace.html
```diff
@@ -1,23 +1,40 @@
...
+                            <button 
+                                class="btn btn-danger" 
+                                ng-click="deleteExceptionById(request.full_stack_trace_id)">Delete</button>
+                        </div>
+                        <div class="d-flex flex-column p-1">
+                            <button 
+                                class="btn btn-secondary" 
+                                ng-click="collapseAllDetails(request.full_stack_trace_id)">Collapse</button>
+                        </div>
+                    </div>
```
NatalieClaraPetersen, line 30: Is there a pop up to confirm your choice before deleting? Where is it, just for curiosity :)

flask_monitoringdashboard/static/pages/exception_stack_trace.html
```diff
@@ -1,23 +1,40 @@
...
+                </div>
             </div>
             <table class="table table-bordered table-hover" width="100%" style="margin: 0;" cellspacing="0">
                 <tbody>
                     <tr ng-repeat="row in request.full_stacktrace" ng-init="key = getUniqueKey(row.function_definition_id, request.full_stack_trace_id)">
                         <td style="padding-left: 10px;">
-                            <details ng-click="getFunctionById(row.function_definition_id, request.full_stack_trace_id)">
+                            <details ng-click="getFunctionById(row.function_definition_id, request.full_stack_trace_id)" id="details_{{request.full_stack_trace_id}}">
```
NatalieClaraPetersen, line 37: we have already defined "key" from getUniqueKey which you can use as the id, or maybe getUniqueKey should just be full_stack_trace_id or maybe even request_id? I think we should align

NatalieClaraPetersen (changes requested):

flask_monitoringdashboard/static/pages/exception_stack_trace.html
AlbertRossJoh: Yes there is, you can use the `confirm` javascript method, which causes the browser to create a pop-up.

flask_monitoringdashboard/static/pages/exception_stack_trace.html
AlbertRossJoh: I disagree, I only use it one place, the use case is very limited, such that we do not need the id multiple times.

Carmish (approved):

PR #485, opened 2025-02-25
removed test that verifies the path to static resources. 

mircealungu:
the tests in this file fail after we removed the generated code from version control. we could ensure that webpack is run before the tests... but still, would that make sense?

i don't think this qualifies as a unit test... it's more testing the capability of flask to serve static resources. 

so i propose that we delete the test.


AlbertRossJoh (approved):
LGTM 🏅

NatalieClaraPetersen (approved):

Carmish (approved):

PR #488, opened 2025-03-03
unique id in frontend at stack lines in recursive functions

NatalieClaraPetersen:
fix so that id in frontend is unique at ExceptionStackLines in recursive functions


AlbertRossJoh (approved):
LGTM 🥇

PR #489, opened 2025-03-03
Cleanup of the exn-monitoring branch

AlbertRossJoh:


flask_monitoringdashboard/database/function_definition.py
```diff
@@ -40,10 +41,9 @@ def get_function_startlineno_and_relativelineno_from_function_definition_id(sess
                     .filter(ExceptionStackLine.function_definition_id == function_defintion_id)
                     .filter(ExceptionStackLine.full_stack_trace_id == full_stack_trace_id)
                     .first())
-    
-    if result is not None:
+    if result is not None and isinstance(result.code, CodeLine):
         return result.code.line_number, result.relative_line_number
     else:
-        return None, None
+        return None
```
Carmish, line 47: The description should be changed as we return None instead of (None, None) if no data is found

Carmish:
LGTM

flask_monitoringdashboard/core/exception_logger.py
```diff
@@ -60,4 +63,4 @@ def log(self, request_id: int, session):
         
         exc_msg_id = add_exception_message(session, str(self.value))
         exc_type_id = add_exception_type(session, self.type.__name__)
-        add_exception_info(session, request_id, trace_id, exc_type_id, exc_msg_id)
+        _ = add_exception_info(session, request_id, trace_id, exc_type_id, exc_msg_id)
```
NatalieClaraPetersen, line 66: instead of adding _ =, we should update add_exception_info to be void, as most other "add_" functions do not have a return value, only those where an ID is needed

NatalieClaraPetersen (changes requested):

flask_monitoringdashboard/core/exception_logger.py
NatalieClaraPetersen: And the return value is never used

PR #493, opened 2025-03-04
Feature/exn monitoring

NatalieClaraPetersen:
Add exception monitoring to the FMD
- A dashboard for a general overview of the exceptions
- An endpoint specific exception dashboard


flask_monitoringdashboard/core/exception_logger.py
```diff
@@ -0,0 +1,108 @@
...
+from flask_monitoringdashboard.database import CodeLine, FunctionDefinition
+from flask_monitoringdashboard.database.exception_info import add_exception_info
+from flask_monitoringdashboard.database.stack_trace_snapshot import add_stack_trace_snapshot, get_stack_trace_by_hash
+from flask_monitoringdashboard.database.exception_stack_line import add_exception_stack_line
+from flask_monitoringdashboard.database.function_definition import add_function_definition
+from flask_monitoringdashboard.database.exception_message import add_exception_message
+from flask_monitoringdashboard.database.exception_type import add_exception_type
+
+
+def hash_helper(s: str):
```
mircealungu, line 19: the fact that it's a helper function is better denoted by prefixing the name with a `_`. Then we can simply call it `_hash` I guess?

flask_monitoringdashboard/core/exception_logger.py
```diff
@@ -0,0 +1,108 @@
...
+    return hashlib.sha256(s.encode('utf-8')).hexdigest()
+
+def h_chain(h: str, tb: Union[TracebackType, None]):
```
mircealungu, line 22: this could be then `_hash_traceback_object`?

flask_monitoringdashboard/core/exception_logger.py
```diff
@@ -0,0 +1,108 @@
...
+    c_line = CodeLine()
+    c_line.filename = frame.f_code.co_filename
+    c_line.line_number = lineno
+    c_line.function_name = frame.f_code.co_name
+    code_context = inspect.getframeinfo(frame).code_context
+    if code_context is not None and len(code_context) > 0:
+        c_line.code = code_context[0]
+    return c_line
+
+def hash_stack_trace(exc, tb):
```
mircealungu, line 47: add small comment here? If i understand correctly the hashed stack trace is the hash of the exception name (?) concatenated to the hash of the traceback object, line by line?

flask_monitoringdashboard/core/user_exception_logger.py
```diff
@@ -0,0 +1,13 @@
+from types import TracebackType
+from typing import Union
+
+ExcInfo = tuple[type[BaseException], BaseException, TracebackType]
+
+class ScopedExceptionLogger():
```
mircealungu, line 6: why the mismatch between the filename and the class name? 
If this represents a `UserCapturedException` then I think that's a better name than Scoped? Otherwise, add a small comment to explain the meaning of `scoped`.

flask_monitoringdashboard/core/measurement.py
```diff
@@ -104,29 +106,40 @@ def evaluate(route_handler, args, kwargs):
...
+
+            return result, status_code
+        except BaseException as e:
+            g.scoped_logger.uncaught_exception_info = sys.exc_info()
+            if isinstance(e, HTTPException):
+                return None, e.code
+            return None, 500
+
+    result, status_code = evaluate_()
+    if len(g.scoped_logger.user_captured_exceptions) != 0 or g.scoped_logger.uncaught_exception_info is not None:
```
mircealungu, line 124: Can we add a little comment on how this exception handling code works at the top of the evaluate method? I can't figure out why `g.scoped_logger.user_captured_exceptions` would ever be non-empty, given that the only thing that we do in evaluate is to assign `uncaught_exception_info`.

mircealungu (changes requested):
a few minor changes and clarification requests.

flask_monitoringdashboard/core/exception_logger.py
```diff
@@ -0,0 +1,153 @@
...
+                # iterate over traceback lines
+                # i.e. the object representation of the following traceback
+                # Traceback (most recent call last):
+                #   File "example.py", line 9, in <module>
+                #     calculate()
+                #   File "example.py", line 6, in calculate
+                #     return divide(10, 0)
+                #   File "example.py", line 2, in divide
+                #     return a / b
+                # ZeroDivisionError: division by zero
```
mircealungu, line 75: I do not understand the "gymnastics", and "this trick". Can we separate them in their own function so it becomes clearer what we're doing and what the comment is referring to? Otherwise, can you give another try to explain what's happening? Tak

mircealungu (commented):
a few comments and questions

flask_monitoringdashboard/core/exception_logger.py
NatalieClaraPetersen: We have resolved the gymnastics by replacing exec_info with a BaseException

flask_monitoringdashboard/core/exceptions/exception_logger.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def __init__(self, scoped_logger: ScopedExceptionLogger):
+        self.user_captured_exceptions: list[BaseException] = (
+            scoped_logger.user_captured_exceptions
+        )
+        self.uncaught_exception: Union[BaseException, None] = (
+            scoped_logger.uncaught_exception
+        )
+
+    def _save_to_db(
```
mircealungu, line 42: this method could be static, or could even be a function - it does not refer to the `self` variable at all.

flask_monitoringdashboard/database/stack_trace_snapshot.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    result = (
+        session.query(StackTraceSnapshot)
+        .filter_by(hash=stack_trace_snapshot_hash)
+        .first()
+    )
+    return result
+
+
+def add_stack_trace_snapshot(session: Session, stack_trace_snapshot_hash: str) -> int:
```
mircealungu, line 21: can we call this create please? it's the traditional naming convention for inserting objects into the DB

flask_monitoringdashboard/core/exceptions/exception_logger.py
```diff
@@ -0,0 +1,109 @@
...
+from flask_monitoringdashboard.database.function_location import add_function_location
+from flask_monitoringdashboard.database.file_path import add_file_path
+from flask_monitoringdashboard.database.function_definition import (
+    add_function_definition,
+)
+from flask_monitoringdashboard.database.exception_message import add_exception_message
+from flask_monitoringdashboard.database.exception_type import add_exception_type
+
+
+class ExceptionLogger:
```
mircealungu, line 32: Could we still do a small rename here? Logger is imprecise here because there are expectations of a `logger` that our object here does not satisfy (e.g. outputs to a stream, collects loglines, etc.). I am thinking about an alternative along the lines of `ExceptionCollector` or `ExceptionCapturer` -  I know, this last one is horrible. But I hope you see my point.  It will make reading the code easier for the first-time reader. 

Also, the same observation for the ScopedLogger.

flask_monitoringdashboard/database/stack_trace_snapshot.py
NatalieClaraPetersen: It aligns with the naming throughout the project eg. add_request, add_stack_line, add_outlier,
Should we then rename these as well?

mircealungu: Uff. Wasn't aware of that. You choose, if we rename, we rename it all please.

flask_monitoringdashboard/core/exceptions/exception_logger.py
NatalieClaraPetersen: renamed to ExceptionCollector and ScopedExceptionCollector

flask_monitoringdashboard/controllers/exceptions.py
```diff
@@ -0,0 +1,157 @@
...
+            "endpoint_id": exception.id,
+            "latest_timestamp": exception.latest_timestamp,
+            "first_timestamp": exception.first_timestamp,
+            "count": exception.count,
+        }
+        for exception in get_exceptions_with_timestamps(session, offset, per_page)
+    ]
+
+
+def delete_exceptions_via_stack_trace_snapshot_id(
```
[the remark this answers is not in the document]
NatalieClaraPetersen: we can remove the intermediate function, do you want us to rename delete_exception_group?

mircealungu (approved):
nice work!

PR #504, opened 2025-03-25
Catch/store uncaught_exception as a BaseException

NatalieClaraPetersen:
catch the exception instead of extracting with exc_info()

### DIFF
--- a/flask_monitoringdashboard/core/exceptions/exception_logger.py
+++ b/flask_monitoringdashboard/core/exceptions/exception_logger.py
@@ -3,7 +3,6 @@ from typing import Union
 
 from sqlalchemy.orm import Session
 from flask_monitoringdashboard.core.exceptions.scoped_exception_logger import (
-    ExcInfo,
     ScopedExceptionLogger,
 )
 from flask_monitoringdashboard.core.exceptions.stack_frame_parsing import (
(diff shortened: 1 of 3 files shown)


flask_monitoringdashboard/core/exceptions/exception_logger.py
```diff
@@ -97,22 +87,30 @@ def save_to_db(self, request_id: int, session: Session):
...
             self._save_to_db(
                 request_id,
                 session,
                 e,
                 type(e),
-                self.uncaught_exception_traceback.tb_next,
+                e.__traceback__.tb_next
             )
+            
+    def get_copy_of_uncaught_exception(self):
```
AlbertRossJoh, line 109: perhaps we want it the other way around, we should "freeze" the exception objects when they enter the scoped logger and perhaps save the original object separetly, such that we can reraise it later?

AlbertRossJoh (changes requested):

flask_monitoringdashboard/core/exceptions/exception_logger.py
NatalieClaraPetersen: Good idea, I fixed it and it solved one of our problems

flask_monitoringdashboard/core/measurement.py
```diff
@@ -119,34 +118,35 @@ def evaluate_():
...
-            return None, 500
+                return None, e.code, e
+            return None, 500, e
 
-    result, status_code = evaluate_()
+    result, status_code, exception = evaluate_()
     if (
         len(g.scoped_logger.user_captured_exceptions) != 0
-        or g.scoped_logger.uncaught_exception_info is not None
+        or g.scoped_logger.uncaught_exception is not None
```
AlbertRossJoh, line 132: Could we not just check if the exception is none?

AlbertRossJoh (approved):

flask_monitoringdashboard/core/exceptions/exception_logger.py
```diff
@@ -97,22 +89,15 @@ def save_to_db(self, request_id: int, session: Session):
...
-            e = self.uncaught_exception_info
-            # We have to choose the next frame as else it will include the evaluate function
-            self._save_to_db(
-                request_id,
-                session,
-                e,
-                type(e),
-                self.uncaught_exception_traceback.tb_next,
-            )
+        # Uncaught Exception
```
AlbertRossJoh, line 96: This comment does not really make sense 😆

AlbertRossJoh (changes requested):

AlbertRossJoh (commented):
LGTM 🎖️

PR #506, opened 2025-04-07
function code is the only thing that is fetched separately

NatalieClaraPetersen:
function code is the only thing that is fetched separately when unfolding details about lines in the stack trace

removed a lot of divs from frontend but it still looks exactly the same

### DIFF
--- a/flask_monitoringdashboard/controllers/exceptions.py
+++ b/flask_monitoringdashboard/controllers/exceptions.py
@@ -121,10 +121,8 @@ def get_function_definition_code(session: Session, function_definition_id: int):
     :param exception_frame_id: the id of the exception frame
     :return: entire code of the function (str)
     """
-    function_code = get_function_definition_code_from_id(
-        session, function_definition_id
-    )
-    return function_code
+
+    return get_function_definition_code_from_id(session, function_definition_id)
 
 
 def _get_relative_file_path_if_in_app(file_path: str):
--- a/flask_monitoringdashboard/controllers/exceptions.py
+++ b/flask_monitoringdashboard/controllers/exceptions.py
@@ -98,6 +98,7 @@ def get_exception_groups_with_details_for_endpoint(
                         exceptionStackLine.path
                     ),
                     "line_number": exceptionStackLine.line_number,
+                    "function_start_line_number": exceptionStackLine.function_start_line_number,
                     "function_name": exceptionStackLine.name,
                     "function_definition_id": exceptionStackLine.function_definition_id,
                     "position": exceptionStackLine.position,
@@ -115,22 +116,17 @@ def get_exception_groups_with_details_for_endpoint(
         )
     ]
 
-def get_exception_function_info(session: Session, exception_frame_id: int):
+def get_function_definition_code(session: Session, function_definition_id: int):
     """
     Retrieves the source code of the function where an exception occurred, the starting line number of the function in the source file, and the relative line number of the exception.
     :param session: session for the database
     :param exception_frame_id: the id of the exception frame
     :return: a dict containing:
-             - start_line_number (int)
              - code (str)
-             - exception_line_number (int)
     """
-    function_start_line_number, exception_relative_line_number, function_definition_id = get_function_info_from_exception_frame_id(session, exception_frame_id)
     function_code = get_function_definition_code_from_id(session, function_definition_id)
     return {
-        "start_line_number": function_start_line_number,
-        "code": function_code,
-        "exception_line_number": exception_relative_line_number
+        "code": function_code
     }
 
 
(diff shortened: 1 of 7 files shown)


flask_monitoringdashboard/controllers/exceptions.py
```diff
@@ -115,23 +113,18 @@ def get_exception_groups_with_details_for_endpoint(
...
-    function_code = get_function_definition_code_from_id(session, function_definition_id)
-    return {
-        "start_line_number": function_start_line_number,
-        "code": function_code,
-        "exception_line_number": exception_relative_line_number
-    }
+    function_code = get_function_definition_code_from_id(
+        session, function_definition_id
+    )
+    return function_code
```
AlbertRossJoh, line 127: I do not really see why we use a variable here

flask_monitoringdashboard/frontend/js/controllers/endpointException.js
```diff
@@ -40,15 +40,13 @@ export function EndpointExceptionController(
             });
     };
 
-    $scope.getUniqueKey = function (exception_frame_id, stack_trace_snapshot_id, stack_trace_position) {
-        return `code_${exception_frame_id}_${stack_trace_snapshot_id}_${stack_trace_position}`; // the row_index is important when dealing with recursive functions
+    $scope.getUniqueKey = function (function_definition_id, stack_trace_snapshot_id, stack_trace_position) {
+        return `code_${function_definition_id}_${stack_trace_snapshot_id}_${stack_trace_position}`; // the stack_trace_position is important when dealing with recursive functions
```
AlbertRossJoh, line 44: Why are we using the `stack_trace_position` for keys?
I believe `function_definition_id` and `stack_trace_snapshot` should be enough?

flask_monitoringdashboard/static/pages/exceptions.html
```diff
@@ -7,39 +7,26 @@ <h4>{{ title }}</h4>
...
-                                request.first_timestamp | dateDifference }}
-                            </h6>
-                        </div>
+                    <div>
+                        <h6>{{ request.message }}</h6>
+                        <h6>
+                            Latest: {{ request.latest_timestamp | dateDifference }} 
+                            | First: {{ request.first_timestamp | dateDifference }}
+                        </h6>
                     </div>
```
AlbertRossJoh, line 20: Why has the classes here been removed?

flask_monitoringdashboard/views/exception.py
```diff
@@ -67,23 +67,21 @@ def get_detailed_exception_info_endpoint(endpoint_id: int, offset: int, per_page
 
 
 @blueprint.route(
-    "/api/function_definition/<int:exception_frame_id>"
+    "/api/function_code/<int:function_definition_id>"
 )
 @secure
-def get_function_definition_for_exception(
-    exception_frame_id
+def get_code_for_function_definition(
```
AlbertRossJoh, line 73: the route and the function name do not match

AlbertRossJoh (changes requested):
Nice changes with some cleanup please adjust the endpoint function, and consider whether we should remove the codeline code in this PR.

flask_monitoringdashboard/frontend/js/controllers/endpointException.js
NatalieClaraPetersen: i removed function_definition_id, but will use it to avoid fetching the same function over and over if recursion

flask_monitoringdashboard/static/pages/exceptions.html
NatalieClaraPetersen: they were unnecessary, it looks exactly as before but with less divs,

flask_monitoringdashboard/frontend/js/controllers/endpointException.js
```diff
@@ -72,8 +70,8 @@ export function EndpointExceptionController(
         }
     };
 
-    $scope.collapseAllDetails = function (name) {
-        document.querySelectorAll(`#details_${name}`).forEach((details) => {
+    $scope.collapseDetails = function (stack_trace_snapshot_id) {
```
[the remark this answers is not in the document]
NatalieClaraPetersen: no it is the details that are queried, I think it is good as is

flask_monitoringdashboard/core/exceptions/stack_frame_parsing.py
```diff
@@ -12,14 +12,3 @@ def get_function_definition_from_frame(frame: FrameType) -> FunctionDefinition:
...
-
-def create_codeline_from_frame(frame: FrameType, lineno):
-    c_line = CodeLine()
-    c_line.filename = frame.f_code.co_filename
-    c_line.line_number = lineno
-    c_line.function_name = frame.f_code.co_name
-    code_context = inspect.getframeinfo(frame).code_context
-    if code_context is not None and len(code_context) > 0:
-        c_line.code = code_context[0]
-    return c_line
```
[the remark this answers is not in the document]
NatalieClaraPetersen: It is okay this time

flask_monitoringdashboard/frontend/js/controllers/endpointException.js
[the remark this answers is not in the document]
NatalieClaraPetersen: we found a middle ground and updated it

PR #513, opened 2025-04-28
rename ExceptionLogger and ScopedExceptionLogger to ExceptionCollector and ScopedExceptionCollector

Carmish:
Rename ExceptionLogger and ScopedExceptionLogger.
Logger is imprecise here because there are expectations of a logger that our objects here does not satisfy (e.g. outputs to a stream, collects loglines, etc.)


NatalieClaraPetersen (approved):
Looks good!

Issue #519, opened 2025-05-15
dashboard Overview Not showing anything its completly empty


gaurav2103:
<img width="953" alt="Image" src="https://github.com/user-attachments/assets/448e2777-3fa8-4b94-a7b2-00fc5bd7c9a6" />

The above screenshot is of console on chrome browser which i am getting after installing the package and running it 

<img width="796" alt="Image" src="https://github.com/user-attachments/assets/1f104c29-2c7a-4cf5-aff5-42e2972f7cec" />

AlbertRossJoh:
@gaurav2103 Would it be possible for you to provide us with information regarding the version of the FMD you're using, and some source code snippet of how you bind the dashboard to the application?

gaurav2103:
> [@gaurav2103](https://github.com/gaurav2103) Would it be possible for you to provide us with information regarding the version of the FMD you're using, and some source code snippet of how you bind the dashboard to the application?

below is the screenshot of FMD version i am using 

<img width="209" alt="Image" src="https://github.com/user-attachments/assets/ea78aead-a6ac-4b50-98f1-72b142b1a4cf" />

and here is the snippet about how i am binding the FMD:

<img width="410" alt="Image" src="https://github.com/user-attachments/assets/3107398a-07b7-44b9-87c2-59541dc7434a" />

@AlbertRossJoh i hope this helps you if it doesn't please let me know what more can i do.

gaurav2103:
@AlbertRossJoh @mircealungu issue is resolved. The issue is already reported and solution was provided in #435. It working fine now for me. The only thing that i am seeing is telemetry and feedback form constantly appearing although i accept it.

<img width="758" alt="Image" src="https://github.com/user-attachments/assets/9a24da2c-8e01-45e1-b037-71209b15ef90" />

AlbertRossJoh:
> [@AlbertRossJoh](https://github.com/AlbertRossJoh) [@mircealungu](https://github.com/mircealungu) issue is resolved. The issue is already reported and solution was provided in [#435](https://github.com/flask-dashboard/Flask-MonitoringDashboard/issues/435). It is working fine now for me. The only thing that i am seeing is telemetry and feedback form constantly appearing although i accept it.

Glad you figured it out! I have made a [PR](https://github.com/flask-dashboard/Flask-MonitoringDashboard/pull/520) informing the bind placement.

[...]

PR #520, opened 2025-05-18
Update README.md

AlbertRossJoh:
Multiple people have experienced issues with placing the bind command, which is solved by placing the bind command after all endpoint definitions(https://github.com/flask-dashboard/Flask-MonitoringDashboard/issues/435) (https://github.com/flask-dashboard/Flask-MonitoringDashboard/issues/519). Therefore I suggest that we hide gif on the front page, until a new one is created and add a comment in the code snippet explaining where the bind command should be placed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,10 @@ A dashboard for automatic monitoring of <a href="https://flask.palletsprojects.c
   <a href="#license">License</a>
 </p>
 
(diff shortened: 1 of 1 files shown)


mircealungu (approved):
Thanks @AlbertRossJoh! This is a good idea!

PR #522, opened 2025-05-21
Fix issue where users can neither accept nor deny telemetry

NatalieClaraPetersen:
Correctly initialize TelemtryController in app.js

There was only an endpoint for checking if telemetry data was "answered" whether yes or no.
Not an endpoint for checking if the answer was "accepted" or not, so if it said "agree" under '/dashboard/configuration' it just indicated that you had clicked either yes or no, but the mechanism for clicking "decline" still worked.

So fortunately nobody gave away telemetry data without consent


mircealungu (approved):
Works well! 
Thanks @NatalieClaraPetersen !

Issue #521, opened 2025-05-20
datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version


aknopper:
It looks to be called directly in these files

/.venv/lib64/python3.12/site-packages/flask_monitoringdashboard/core/timezone.py:13: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    return dt + config.timezone.utcoffset(datetime.datetime.utcnow())
 
/.venv/lib64/python3.12/site-packages/flask_monitoringdashboard/core/cache.py:80: DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).
    memory_cache.get(endpoint_name).set_last_requested(datetime.datetime.utcnow())

aknopper:
@mircealungu can you please look at this? I believe I have the fix in this PR

PR #527, opened 2025-05-28
Fix UTC Timezone

aknopper:
datetime.datetime.utcnow() is deprecated in favor of datetime.datetime.now(datetime.UTC)

Fixed where this is referenced in the code.


mircealungu (approved):
nice. thanks @aknopper aknopper

figgeous:
Closes #521 ?

PR #537, opened 2025-10-28
Change utcnow to now(datetime.timezone.utc)

aknopper:
-Found more instances of the deprecated utcnow function and changed to .now(datetime.UTC)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -5,6 +5,23 @@ All notable changes to this project will be documented in this file.
 This project adheres to `Semantic Versioning <http://semver.org/>`_.
 Please note that the changes before version 1.10.0 have not been documented.
 
+v5.0.2
+----------
+**Bug Fix Release**
+
+Fixed
+^^^^^
+- **Build**: Fixed broken 5.0.0 release - Angular frontend is now properly built before publishing to PyPI
+- **Deprecation**: Replaced deprecated ``datetime.utcnow()`` with ``datetime.now(timezone.utc)`` throughout codebase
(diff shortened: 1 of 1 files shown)


flask_monitoringdashboard/core/database_pruning.py
```diff
@@ -23,7 +23,7 @@
 def prune_database_older_than_weeks(weeks_to_keep, delete_custom_graph_data):
     """Prune the database of Request and optionally CustomGraph data older than the specified number of weeks"""
     with session_scope() as session:
-        date_to_delete_from = datetime.utcnow() - timedelta(weeks=weeks_to_keep)
+        date_to_delete_from = datetime.now(datetime.timezone.utc) - timedelta(weeks=weeks_to_keep)
```
mircealungu: @aknopper - you're sometimes using `datetime.UTC` and sometimes `datetime.timezone.utc` as you do here. the best would be actually to do `datetime.timezone.utc`. the former is only available in Python 3.11. Could you hunt all the occurrences and change them? Also from the past PR.

aknopper: @mircealungu done!

flask_monitoringdashboard/views/reporting.py
```diff
@@ -19,7 +19,7 @@
 
 
 def get_date(p):
-    return datetime.utcfromtimestamp(int(request.args.get(p)))
+    return datetime.fromtimestamp(int(request.args.get(p)),tz = datetime.datetime.now(datetime.timezone.utc))
```
mircealungu: one small change: the `tz=...` expects a timezone object not a datetime. can you fix this?

mircealungu: ok. no worries; i fixed it myself.

mircealungu (approved):
looks good.

PR #544, opened 2025-11-26
quick fix for the timezone issue for issue 543

nicfar1708:


mircealungu:
Dear @nicfar1708,

  Thanks for identifying the problem and proposing a fix and sorry for the slow response time!

I think there's a simpler and more general solution: instead of fixing the insertion, we could fix the model defaults to use a callable:

```python
  # Instead of (evaluated once at import):
  default=datetime.datetime.now(datetime.timezone.utc)

  # Use (evaluated each time):
  default=lambda: datetime.datetime.now(datetime.timezone.utc)
```

  There are a few more instances of the problem than the ones you identified (you can find them by searching for `default=datetime`).

  Would you like to try this fix?

nicfar1708:
Hello @mircealungu,

thank you for the feedback.
I'm aware of the _lambda_ solution but, personally, I prefer to be more "explicit" when writing code (even if it is less pythonic).

I will submit a fix for this in the following days by using your proposed solution.

waldekwg (approved):
We confirm the bug and the proposed solution

waldekwg:
@aknopper The bug was introduced with the last release [fix timezones to UTC](https://github.com/flask-dashboard/Flask-MonitoringDashboard/commit/0741e687bd859b0b7a7dc638c67c7b4635d07670)

The previous version was correct
`time_added = Column(DateTime, default=datetime.datetime.utcnow)`
in the new version it's initialized only once
`time_added = Column(DateTime, default=datetime.datetime.now(datetime.timezone.utc))`

Unfortunately, it broke the user statistics for the past few months :-/

What's the fastest way to commit the fix and deploy a new version of the library?

